PR #1244, opened 2012-08-17
Crud generation composite primary key

edisnake:
Hi with this changes yii will can generate crud code for database tables with composite primary keys, checked it out

### DIFF
--- a/framework/gii/generators/crud/CrudCode.php
+++ b/framework/gii/generators/crud/CrudCode.php
@@ -66,8 +66,6 @@ class CrudCode extends CCodeModel
 			$table=CActiveRecord::model($class)->tableSchema;
 			if($table->primaryKey===null)
 				$this->addError('model',"Table '{$table->name}' does not have a primary key.");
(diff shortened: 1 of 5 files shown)


samdark:
Can't be merged as is and it seems it's not going anywhere. Closing.

teo1978:
So, there have been 3 pull requests attempting to offer support for CRUD for composite primary keys (a functionality whose lack has always been astonishing in the first place), all have been closed without success, and we still don't have CRUD for composite primary keys? Do I understand the situation correctly?

samdark:
@teo1978 something like that. Do you want to try?

teo1978:
I don't think so.
Has it at least been implemented in Yii2?

samdark:
Try it.

nineinchnick:
I added support for it in Yii2 in my custom gii generator, it requires methods to encode the pk values as strings.

cebe:
Yii2 gii has it implemented.

Issue #218, opened 2012-02-15
no order by in union select


qiangxue:
In CDbCommand class function buildQuery()

has the following functionality:

if(isset($query['order']))
  $sql.=&quot;\nORDER BY &quot;.$query['order'];

  $limit=isset($query['limit']) ? (int)$query['limit'] : -1;
  $offset=isset($query['offset']) ? (int)$query['offset'] : -1;

  if($limit&gt;=0 || $offset&gt;0)
   $sql=$this-&gt;_connection-&gt;getCommandBuilder()-&gt;applyLimit($sql,$limit,$offset);
## if(isset($query['union']))

As You see Order by condition goes before union statement, so there are anny oportunity to gave Order by condition to union select, for example:

(Select id, title from test)
UNION
(Select id, title from test2)
Order by title DESC

This is not the same as gave an order by condition to each of the selects.

Good luck

<p>Migrated from http://code.google.com/p/yii/issues/detail?id=2682</p><hr/><h2>earlier comments</h2>
<p><strong>qiang.xue said, at 2012-01-01T03:37:10.000Z:</strong></p>

<p>set for 1.1.10 milestone</p><p><strong>qiang.xue said, at 2012-01-01T03:37:35.000Z:</strong></p>

<p>set for 1.1.10 milestone</p>

PR #1788, opened 2012-12-01
#218 sql builder union fix

nsanden:
i did some basic testing, seems like it works fine.

Fix for #218


alexeygon:
What about `LIMIT`? Same logic need

samdark:
@alexeygon please submit separate issue for it.

PR #1876, opened 2012-12-21
CActiveRecord type casting

pavlepredic:
This is an attempt to solve the issue of fetching data of incorrect type from DB (eg fetching integer as string). The proposed solution will perform type-casting when populating record, according to specifications in AR meta data. Default behavior is NOT to perform type casting. User will need to specify in each AR model that he wants to use type casting.


samdark:
Won't implement it in 1.1.

PR #2556, opened 2013-06-12
Namespace in the component configuration

petrgrishin:
Hello!

I use in projects Composer to manage dependencies vendors and also for **autoload classes**.

I can use the namespace to create yii application:

``` php
Yii::createApplication('\MyNamespace\WebApplication', $config)->run();
```

But can not use the namespace in the component configuration:

``` php
'behaviors' => array(
     'mybehavior' => array(
         'class' =>  '\MyNamespace\Class',
     ),
 ),
```

I can not find a more elegant solution .. May have any thoughts?


petrgrishin:
:+1:

PR #2687, opened 2013-07-26
Issue #2653: updated Markdown Extra Lib to version 1.3

SleepWalker:


SleepWalker:
@hightman, @samdark probably it is too late, but the problem was because of different structure of `$matches` in `_MarkdownExtra_TmpImpl::_doFencedCodeBlocks_callback($matches)`. I don't know how it was working by me earlier, but now it is definitely not (I occasionally plugged the modified version of Yii in my current project and found out this bug...).
Here is example from 1.3:
https://github.com/yiisoft/yii/blob/dcec697c06cfc1ffd3591586071644fc72ffda36/framework/vendors/markdown/Markdown.php#L2797
And here is current markdown version from Yii:
https://github.com/yiisoft/yii/blob/1.1.16/framework/vendors/markdown/markdown.php#L2378

So if we change index from 2 to 4 in [this row](https://github.com/yiisoft/yii/blob/1.1.16/framework/utils/CMarkdownParser.php#L113) everything should work fine.

samdark:
@SleepWalker you mean the bug is in current master?

SleepWalker:
No, I don't. Current master works perfectly, because there is old version of Markdown.

samdark:
Well, I don't think we're going to update it for 1.1 unless you pull request again with tests and for 2.0 we're using @cebe implementation that's better.

SleepWalker:
Ok

PR #2894, opened 2013-09-22
Fix problem with spaces in tablenames

devivan:
using a Microsoft SQL server I found table names with spaces O_o.
I develop a patch that, mantaining the back compatibility, match table having spaces requiring the use of an alias (with or without AS).


samdark:
Thanks!

lennartvdd:
The same problem exists for column names. It would be nice to apply this in the select() method as well.

PR #2900, opened 2013-09-23
introduce new emptySummaryText property in CBaseListView

### LINKED ISSUE #2898: Allow CBaseListView to display summary in any conditions
Add new poperty `CBaseListView.summaryAlwaysVisible`, set to `false` by default, and parametrize [line 205 of `CBaseListView.php`](https://github.com/yiisoft/yii/blob/1.1.14/framework/zii/widgets/CBaseListView.php#L205) so it would be executed only, if `CBaseListView.summaryAlwaysVisible=false`.

Currently Yii does not display summary line for `CBaseListView`, if number of items from connected data provider is less or equal zero. This seems to be wrong and such behavior should be decided by a developer, not by a framework iself.

**Yes**, I understand that I can extend `CBaseListView` and add `summaryAlwaysVisible` or change `renderSummary` myself, but **no**, this is wrong idea. This should be done at framework level. Forcing developer to anything is wrong. Let him or her decide.

nineinchnick:
Fixes #2898


mdeweerd:
Seems related with https://github.com/yiisoft/yii/issues/3498 .

PR #2903, opened 2013-09-24
Render "exception" view always if YII_DEBUG is true

DragonBlack:
YII_DEBUG should have the highest priority

### DIFF
--- a/framework/base/CErrorHandler.php
+++ b/framework/base/CErrorHandler.php
@@ -344,7 +344,7 @@ class CErrorHandler extends CApplicationComponent
 	 */
 	protected function renderError()
 	{
-		if($this->errorAction!==null)
+		if($this->errorAction!==null && !YII_DEBUG)
 			Yii::app()->runController($this->errorAction);
 		else
 		{


rob006:
Any progress with this? Maybe something like that:
```php
	/**
	 * Renders the current error information.
	 * This method will display information from current {@link error} value.
	 */
	protected function renderError()
	{
		if($this->errorAction!==null && (!YII_DEBUG || $this->getException() instanceof CHttpException))
			Yii::app()->runController($this->errorAction);
		else
		{
			$data=$this->getError();
			if($this->isAjaxRequest())
				Yii::app()->displayError($data['code'],$data['message'],$data['file'],$data['line']);
			elseif(YII_DEBUG)
				$this->render('exception',$data);
			else
				$this->render('error',$data);
		}
	}
```

samdark:
@yiisoft/core-developers opinions on how to solve it?

realtebo:
@klimov-paul  

I corrected your config code as:

    errorHandler => array(
        'errorAction' => YII_DEBUG ? null : 'site/error'
    )

because on product it give a "Unalbe to resolve request 'Array'" error

samdark:
Closing since it's already part of application template.

rob006:
> Closing since it's already part of application template.

Current solution does not allow for testing error pages (eg 404) on the development environment.

samdark:
It does if you configure it like

```php
'errorAction' => 'site/error'
```

rob006:
Then I will not get stack trace on any other errors. So I need constantly switch config on dev environment to have both - debug stack and nice error pages for 404 errors. 
What is wrong with the solution in https://github.com/yiisoft/yii/pull/2903#issuecomment-71648561 ?  CHttpExceptions does not need stack trace, because they usually do not mean an application error.

samdark:
You solution doesn't allow displaying error action in non-debug mode.

rob006:
I allows - it uses error action when exception is instance of CHttpException.

samdark:
I mean non-HTTP errors.

rob006:
Why do you want display error pages for application errors on dev environment?

samdark:
To develop it custom error handler.

rob006:
If you have custom error handler you can override `renderError()` method and do whatever you want. But current behaviour is pretty annoying for regular development and most common cases. If we really want to have ability to use error action for non-http errors on dev environment we can add additional property for CErrorHandler like `forceErrorActionForErrors` which default will be set to `!YII_DEBUG`.

PR #2911, opened 2013-09-27
Added id parameter to CSort and CPagination

cnlevy:
CDataProvider should not know about pagination->pageVar and
pagination->sortVar. Setting the 'id' param allowed me to use a subclass
of CPagination which also has a pageSizeVar to handle pageSize the same
way as pageVar


samdark:
Won't be fixed in 1.1.

Issue #2936, opened 2013-10-08
STAT relation does not allow custom key associations


louis89:
The STAT relation type does not allow custom key relations like HAS_MANY relation type does. It is sometimes very convenient or necessary to be able to specify a custom keys in a relationship. HAS_MANY relations allow specifying custom keys using the array('FK'=>'PK') syntax, but this syntax does not work for STAT relations.

PR #2937, opened 2013-10-08
Added option to specify custom key associations in STAT relations.

louis89:
Fixes issue #2936

This PR adds the option to specify custom key associations in STAT relations using the syntax array('FK'=>'PK'). This is the same syntax as used by HAS_MANY relations to specify custom key associations. This is an extensions of the current functionality of STAT relation, no changes to previous STAT relation functionality or syntax have been made.


samdark:
Won't be fixed in 1.1.

PR #2952, opened 2013-10-10
added support for pecl redis extension

Andrewpk:
Added support for pecl "redis" extension and additional error/exception handling when using the php redis extension. 

I thought this would be a good idea due to phpredis' availability on both Windows and *nix-based platforms.

I probably should have filed a feature request issue for this first, but I figured it wouldn't take me that long just to implement the feature thanks to Carsten Brandt's framework.

If someone should decide to add phpiredis to the mix, it shouldn't be that difficult as I made sure to leave an opening for additional php redis extensions as they come and go.


samdark:
Closing since 1.1 isn't accepting enhancement anymore.

PR #2956, opened 2013-10-12
CJuiDatepicker fix a bug when defaultOptions are set but 'language' param is not set (Chinese instead of English)

ekonoval:
When no 'language' is set for datepicker widget, but the 'defaultOptions' param is filled, the calendar is displayed in Chinese (or some other language, I'm not quite sure, using the  hieroglyphs) when it must be displayed in English. I encountered with this behaviour in a lot of questions on the internet (especially amoung yii beginners like me).
I'm attaching 2 screenshot with bug and after my fix.

Also here https://www.dropbox.com/s/fkvx4dj8kwnthsz/jui_bug_chinese.zip you can find a small yii view code which can be usefull to reproduce the bug.

bug
fix


samdark:
Won't be fixed in 1.1.

PR #3021, opened 2013-11-07
Yii core views rendered from a non-global namespace

jorgebg:
Now the views can be rendered from a non-global namespace


framework/views/ar/log-firebug.php
```diff
@@ -7,9 +7,9 @@
 foreach($data as $index=>$log)
 {
 	$time=date('H:i:s.',$log[3]).sprintf('%03d',(int)(($log[3]-(int)$log[3])*1000));
-	if($log[1]===CLogger::LEVEL_WARNING)
```
jorgebg: a

PR #3032, opened 2013-11-10
Added wave transforms to captcha

rezakho:


samdark:
Won't be added in 1.1.

PR #3072, opened 2013-11-25
Validation of all the fields when you change one

kikimor:
If logic of form is "field change can affect other fields", you may need to validate all fields, not just the one that was changed.


samdark:
Won't be fixed in 1.1.

Issue #1565, opened 2012-10-14
COutputCache should respect Content-Type headers


navarr:
Right now COutputCache (when used as a filter) does not listen for any headers sent with the page, so far as I can tell.

This means that it can not cache json, images, or other content that might be dynamic without extending upon it.

COuputCache should honor at least the content-type header sent by the page when it is being used as a filter.

martinytodorov:
I just added a new class that will help me with the caching of the RESTful module and its headers:
```
class ApiOutputCache extends COutputCache
{

    public function init()
    {
        if(Get('format') == 'xml')
            $content_type = 'text/xml';
        else
            $content_type = 'application/json';

        header('Content-type: ' . $content_type . ";charset=utf-8");
        return parent::init();
    }

}
```

PR #3117, opened 2013-12-17
Issue #1565: COutputCache should respect Content-Type headers

s12v:
Issue #1565: COutputCache should respect Content-Type headers

Allows Content-Type, Content-Length, Content-Encoding, and Etag headers to be cached


samdark:
Won't be fixed in 1.1.

PR #3118, opened 2013-12-17
Support labels with attribute bracket notation like post[author][name]

cnlevy:
It allows to write the following code in views:

``` php
echo $form->checkBox($model,'params[hasCar]');
echo $form->labelEx($model,'params[hasCar]'); // before, i had to write 'params.hasCar'
echo $form->error($model,'params[hasCar]');
```

It has the additional advantage of correctly generating the 'for' attribute, i.e. the label will be 

``` html
<label for="Offer_params_hasCar">Possède une voiture</label>
```

instead of 

``` html
<label for="Offer_params.hasCar">Possède une voiture</label>
```

which wasn't focusing on the associated input


samdark:
Won't be fixed in 1.1.

PR #3244, opened 2014-02-24
Fix #3243: Correct CPgsqlSchema default column expression and sequence name

### LINKED ISSUE #3243: CPgsqlSchema fails to derive correct default column values
`CPgsqlSchema::findColumns()` makes a query to the database to figure out what columns there are, and also their default value. It gets the default value information from the `adsrc` column in the `pg_attrdef` table. The default value returned from this query may can be incorrect in situations where the default value is e.g. `nextval('some_sequence'::regclass)` and some_sequence has been renamed. Perhaps it can be incorrect in other cases too, this is the one I encountered the issue with.

For example, if we had a table `trip__region` that we renamed to `tour__region`, and in doing so we also renamed the primary key `id`'s corresponding sequence using `ALTER SEQUENCE trip__region_id_seq RENAME TO tour__region_id_seq`, the default value for the `id` column is now (as far as PostgreSQL is concerned) `nextval('tour__region_id_seq'::regclass)` (after previously being `nextval('trip__region_id_seq'::regclass)`).

However, while PostgreSQL knows about and uses the correct default value, the old one is still stored in the (deprecated) `adsrc` column in the `pg_attrdef` table - it is not updated when we rename the sequence. And since that column is what Yii looks at to determine the default value of the column, it gets the wrong expression. For reference: http://www.postgresql.org/docs/current/static/catalog-pg-attrdef.html

Aside from not knowing about the right default expression for the column, this also causes problems because it derives an incorrect sequence name from that expression (resulting in errors for example when using `resetSequence()` with a sequence name based on the (incorrect) schema information.

To fix this, we should use the `pg_get_expr()` function instead of the `adsrc` column. This will return the correct default value expression for the column(s). I'm preparing a PR.

rawtaz:


samdark:
Won't be fixed in 1.1.

rawtaz:
You're giving up on 1.1? :-)

samdark:
http://www.yiiframework.com/news/90/update-on-yii-1-1-support-and-end-of-life/

rawtaz:
R.I.P 1.1 - Easily the best Yii version in terms of nostalgia! 👍

samdark:
We'll release 1.1.18 w/ PHP 7 fixes. That would probably be last release.

PR #3266, opened 2014-03-08
Fix linkLabelWrapperHtmlOptions

mbdwey:
what if i each Wrapper element need it's own html options !


samdark:
Won't be fixed in 1.1.

PR #3320, opened 2014-04-04
Fixed #3303

### LINKED ISSUE #3303: checkBoxList 'container' option and client side validation
By default, checkBoxList adds a 'span' container around checkbox items.

Recently there was a need to change this to match the specific CSS styling for the application.

I needed the container to be `<div class="options">`. So I tried the following:

```
<div class="row">
    <?php echo $form->labelEx($model, 'category'); ?>

    <div class="options">
        <?php echo $form->checkBoxList($model, 'category', $options, array(
            'template' => '<div class="item">{input}{label}</div>',
            'separator' => false,
            'container' => false,
        )); ?>
    </div>

    <?php echo $form->error($model, 'category', array('inputContainer' => '.row')); ?>
</div>
```

This gives the correct layout, however the problem with this is that the attribute does not get marked in error class when there is a validation error (in client side validation).

So the next thing I tried was to set `'container' => 'div class="options"'` (and remove my custom container markup). This actually worked but on close inspection I found that it was also adding `<div class="options">` as the closing tag (instead of just `</div>`).

In the end I figured out how to make this work correctly. I have to create my custom container markup as follows:

`<div class="options" id="User_category">`

The client side validation looks for an element with an ID that is related to the checkboxlist. If you specify a value for `container`, it automatically inserts the ID value but if you specify `false` then since no container is generated it is our responsibility to insert this ID value.

So I think one of the following should be done to resolve this issue:
1. A new option `containerCssClass` to allow us to specify a class name for the container (default `null`), OR
2. Documentation to be updated to advise users how to retain client side error highlighting when using custom container markup.

nukkumatti:
Fixed #3303

### DIFF
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -1229,6 +1229,7 @@ EOD;
 	 * for every label tag in the list.</li>
 	 * <li>container: string, specifies the radio buttons enclosing tag. Defaults to 'span'.
 	 * If the value is an empty string, no enclosing tag will be generated</li>
+	 * <li>containerCssClass: string, specifies the checkboxes enclosing tag's class.</li>
 	 * <li>baseID: string, specifies the base ID prefix to be used for radio buttons in the list.
 	 * This option is available since version 1.1.13.</li>
 	 * <li>empty: string, specifies the text corresponding to empty selection. Its value is empty.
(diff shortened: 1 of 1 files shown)


samdark:
Won't be fixed in 1.1.

Issue #3103, opened 2013-12-10
backslash "\" not treated correctly by createUrl


teo1978:
Sometimes you are forced to escape some special characters with a backslash in the regular expression used in a urlManager rule if you need to match those characters literally, for example:

    'foo/\*<areas:[^\d].*>' => 'foo/bar'
where I use \* to match a literal "*" in the url.

These url rules will be processed correctly by parseUrl(), but not by createUrl()

So, in the example above, an url like '/foo/*hello' will be correctly parsed as equivalent to /foo/bar?areas=hello

However, Yii::app()->createUrl('foo/bar', 'areas'=>'hello') will output "/foo/\\\*hello" instead of the expected "/foo/*hello", that is, it puts an extra backslash. 

In turn, some browsers (namely Chrome) will automagically transform that to /foo//*hello, for some reason, i.e. replacing the "\" with a "/", which I don't know whether it is standard-compliant or a browser's bug, but that's pretty irrelevant here.

lierdakil:
So do you suppose this can be closed now?

qiangxue:
Thanks for your help!

eirikhm:
This actually introduced a regression, see #4051

PR #3348, opened 2014-05-01
Fix for #3103

lierdakil:
Fix for #3103
Applies preg_quote to every string between angle-bracket parameters
while building CUrlRule::pattern if CUrlRule::parseOnly==false.
This allows using non-escaped perl regex special characters in
url patterns.


lierdakil:
Sooo... anything I can do to expedite this process? For a simple patch like this, nine months feel like eternity...

samdark:
@lierdakil yes. It's a central part of the framework so I'm a bit worried about how it may affect performance.

lierdakil:
As I said earlier, with caching it won't, since it will only be executed once for a given ruleset. As for scenarios without caching, I'd say that in such scenarios performance is not a primary concern. But if you want, I could run some tests, if you point me where to start (profiling php is not my forte, to tell the truth, so a little advice would be helpful)

samdark:
1. Create a simple app with 30 url rules.
2. Make a request so the last one matches (i.e. all 30 rules should be matched).
3. Measure it before/after patch.

If there's no significant difference, everything's OK.

lierdakil:
Okay, some test results:
Mean generation time from 1000 trials, as reported by `CLogger::getExecutionTime()`
#### No caching
Without patch: 0.0137±0.0004 s
With patch: 0.0139±0.0005 s
Difference is well inside error margin.
#### File cache
Without patch:  0.0134±0.0003 s
With patch: 0.0133±0.0005 s
Difference is well inside error margin.
#### Conclusion
No significant difference found, measured differences are superficial and can safely be attributed to measurement errors/random perturbations.

samdark:
Thanks! @qiangxue I think it's safe to merge it but since it's pretty much central thing it would be great if you'll check the code as well.

framework/web/CUrlManager.php
```diff
@@ -689,7 +695,10 @@ public function __construct($route,$pattern)
 		$this->append=$p!==$pattern;
 		$p=trim($p,'/');
 		$this->template=preg_replace('/<(\w+):?.*?>/','<$1>',$p);
-		$this->pattern='/^'.strtr($this->template,$tr).'\/';
+		$p=$this->template;
+		if(!$this->parsingOnly)
+			$p=preg_replace_callback('/(?<=^|>)[^<]+(?=<|$)/',array($this,'escapeRegexpSpecialChars'),$p);
```
qiangxue: Why is this only called when `!$this->parsingOnly`?

framework/web/CUrlManager.php
```diff
@@ -638,6 +638,13 @@ class CUrlRule extends CBaseUrlRule
 	public $hasHostInfo;
 
 	/**
+	 * Callback for preg_replace_callback in counstructor
+	 */
+	protected function escapeRegexpSpecialChars($matches) {
```
qiangxue: Needs to adjust code style...

framework/web/CUrlManager.php
lierdakil: > Fixed the bug that CUrlRule treated special regex characters differently in createUrl and parseUrl

Because it *might* be instrumental to use regex in parsing only url rules. Problem was that createUrl and parseUrl behaved differently. Since using createUrl with parsing-only rules is impossible (so bug does not manifest there), and using regex wildcards in urlRules is a valid option, I used this conditional. It's reasonable when you think about it.

framework/web/CUrlManager.php
lierdakil: I assume you mean newline before `{`?

lierdakil:
@qiangxue, I hope I addressed your concerns. If not, please point me to what needs work still.

qiangxue:
Thanks!

PR #3387, opened 2014-05-26
[RFC] Enabling validation that depends on custom ajax

igorsantos07:
This includes a setInterval on yiiactiveform.validate so it can wait for all validations that depends on async functions (such as AJAX callbacks) to finish properly.

Backwards compatibility is maintained, but if there's any custom validation plugin that depends on async callbacks to consider the field is valid or not, it can use a new property called `attribute.validated`. Setting it to `false` will hold validation until it's set to `true` again.

Example:

``` js
attribute.validated = false; //this will make the main function verify if the field finished validation again
var val = myHandler(attribute.value); //this will work on the value and get something else from it
$.ajax({
    url: 'myvalidationserver.com',
    data: { value: val },
    success: function(data) {
        if (!data.valid) {
            messages.push('Boo, invalid stuff!');
        }
        attribute.validated = true; //it will finally stop verifying end of validation procedure in the next verification
    }
});
```

Points to evaluate:
- The interval is 100ms which is just for testing purposes; maybe it should be higher to not stop the browser completely?
- I've not added any documentation on the new "validated" property. I have no idea where this should go, since I started writing client-side validation plugins by copying the framework one's. Is there any guide/help/doc page on how to create your custom validations?
- I've intentionally used weird identation so we can easily see what code has been added without bothering with whitespace changes. Once (if ever) the PR gets positive comments I'll add another commit that fixes identation, so it can be merged :) Who knows, maybe someday GitHub will have a "ignore whitespace" diff option :P*

### DIFF
--- a/framework/web/js/source/jquery.yiiactiveform.js
+++ b/framework/web/js/source/jquery.yiiactiveform.js
@@ -321,20 +321,40 @@
 			settings = $form.data('settings'),
 			needAjaxValidation = false,
 			messages = {};
+
 		$.each(settings.attributes, function () {
 			var value,
 				msg = [];
+
+			this.validated = true;
+			this.finished = true;
+
 			if (this.clientValidation !== undefined && (settings.submitting || this.status === 2 || this.status === 3)) {
 				value = getAFValue($form.find('#' + this.inputID));
 				this.clientValidation(value, msg, this);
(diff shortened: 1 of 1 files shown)


samdark:
Won't be fixed in 1.1.

PR #3398, opened 2014-06-05
Fixed CJSON::encode's handling of resources

a-square:
Now a resource is encoded as `null`, which is better than having malformed JSON resulting from the default behavior


samdark:
Won't be fixed in 1.1.

Issue #2881, opened 2013-09-18
CGridView sometimes ignores filter change after ajax update


MonkeyMaster:
It happens when CGridView has both text field and dropdown filters.
1. Enter text and press ENTER - grid updates (eventType is now 'keydown')
2. Select an option - grid does not update! ('change' event is ignored because eventType is still set to 'keydown'. after that eventType is changed to '')
3. select another option - grid updates

So eventType must be reset after each grid ajax update.

My solution:
```js
var data = $('#' + id).data();
if (event.type === 'keydown') {
    if (event.keyCode !== 13) {
        return; // only react to enter key
    } else {
        data.eventType = 'keydown';
    }
} else {
    // prevent processing for both keydown and change events
    if (data.eventType === 'keydown') {
        data.eventType = '';
        return;
    }
}
```

PR #3401, opened 2014-06-06
Prevent processing for both keydown and change events on the same element

sivir:
Hello! This fix checks that keydown and change event prevented only on the same element. It fixes bug, that can be reproduced this way:
1. Create filter for to field, one filter must be text input, other must be checkbox
2. Open any version of Firefox
3. Input text for search in text field end press Enter. At this moment this script will set variable eventType to 'keydown'
4. Click on checkbox. Filter won't run, because eventType is 'keydown', even though change event called by another element.
I think, it's related to bug #2881 too.


sivir:
Is there anybody who can merge it?

samdark:
Yes, we can :)

samdark:
Merged https://github.com/yiisoft/yii/commit/8c30239dce366a70bc7486cd605c00c4938f0c9e Thanks!

PR #3504, opened 2014-08-25
Fix #3046, Client validation with skipOnError option

### LINKED ISSUE #3046: Captcha validator bug with client validation
Suppose I have the following rules for captcha validation:

```
array('verifyCode', 'required'),
array('verifyCode', 'captcha', 'allowEmpty'=>!CCaptcha::checkRequirements(), 'skipOnError'=>true),
```

With client side validation turned on (not ajax), the error is still generated (in errorSummary) for the second rule (i.e. it does not seem to respect the `skipOnError` setting).

One other thing is that the error message that gets generated by default does not use the attribute label, i.e. it seems to be hard coded as 'verification code'.

SG5:
Fix for #3046


samdark:
Won't be fixed in 1.1.

PR #3509, opened 2014-08-28
Allow to use media parameter for css files in packages

bizley:
When adding css files to package we can specify media parameter for the file like
'css' => array('first-file-url-no-media', 'second-file-url-with-media'=>'media-like-print-or-whatever')


samdark:
Won't be fixed in 1.1.

PR #3529, opened 2014-09-08
yiigridview.getChecked should return array of checkbox's values instead of keys

dlnsk:
It allow to change 'name' parameter of CCheckBoxColumn.

I use MANY-MANY (Child-ChildSchool-School) relation, so I select records from ChildSchool but display related Childs like this:

```
'columns'      => array(
    array(
        'class' => 'CCheckBoxColumn',
        'id' => 'child-boxes',
        'name' => 'child.id',
        'selectableRows' => 2,
    ),
    array(
        'name' => 'child.lastname',
        'value' => 'CHtml::link(CHtml::encode($data->child->lastname),
                                 array("view","id" => $data->child->id))',
        'type' => 'raw',
        'filter' => CHtml::activeTextField($model, 'lastname'),
    ), 
    etc
```

In this case even if I use 'name' in CCheckBoxColumn yiigridview.getChecked returns keys of ChildSchool instead of 'child.id'.

### DIFF
--- a/framework/zii/widgets/assets/gridview/jquery.yiigridview.js
+++ b/framework/zii/widgets/assets/gridview/jquery.yiigridview.js
@@ -392,14 +392,14 @@
 		 */
 		getChecked: function (column_id) {
 			var settings = gridSettings[this.attr('id')],
-				keys = this.find('.keys span'),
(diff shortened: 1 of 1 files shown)


samdark:
Won't be fixed in 1.1.

PR #3543, opened 2014-09-16
resetDefaultScope option in condition AR

yamaha252:


samdark:
Won't be fixed in 1.1.

PR #3547, opened 2014-09-20
Added extra configuration tag to a li node in CMenu class

avallem:
With this add we handle which tag is rendered inside a li node


samdark:
Won't be fixed in 1.1.

PR #3575, opened 2014-10-08
Set default column class for CGridView

bkwint:
When you currently want to use a custom column class for a column you have to define the class property in the column array definition. When you want to overwrite the column class for every column in the grid view you have to either define the class property for every column or you have to extend the CGridView with your own implementation and overwrite the two functions createDataColumn($text) and initColumns().

This implementation of the CGridView helps you to make this more easy. A property has been added to the CGridView named $defaultColumnClass. This property has the default value CDataColumn but this can be overwritten when setting up your grid view. When you overwrite this any column that does not have a class defined in it, will automatically use the newly given DataColumn class.


samdark:
Won't be fixed in 1.1.

PR #3590, opened 2014-10-13
fix #3589 using the proposed tactic.

### LINKED ISSUE #3589: CJSON::decode does not properly flag invalid JSON input as invalid (return NULL)
When the bad content resides inside an array in the JSON input string, then that array content is decoded as NULL but the CJSON::decode() call itself still produces the rest and thus acts as if the input is legal (valid).

Minimal example to show the difference between PHP json_decode() (PHP 5.4+) and CJSON::decode() -- note the incorrect "y" array element: the colon is illegal --:

```
var_dump(json_decode('{ "x": [ "y" : { "A": 1 }], "z": 7 }'));
//   => NULL 

var_dump(CJSON::decode('{ "x": [ "y" : { "A": 1 }], "z": 7 }'));
//   => array(2)
//      x => array(1)
//          0 => NULL 
//      z => integer 7
```

The reason CJSON::decode() fails is because apparently decode errors are not propagated in its own PHP-based decoder, which, thanks to #403, is invoked when PHP native `json_decode()` returns NULL, as it does here.
## I have not provided a fix, because...:

Looking at the CJSON.php code, you'll find that the place where the array is parsed at line 515:

```
$arr[] = self::decode($slice,$useArray);
```

cannot easily detect where the returned NULL is a 'proper' NULL (because the input `$slice` was `null`) or an error-signaling NULL.

One option would be to pass along an variable reference which can be used to flag the error in any recursive call to `self::decode()`. Or one could throw an exception and catch it at the root call level. 
I don't know which solution flavor is prefered by the YII coders (personally I would go for the extra pass-by-reference parameter + if() check, assuming that would produce generally faster (as in: less slow) code in PHP, compared to nested try/catch exception stacks). 

**Side Note**: given the way CJSON::decode() is written, I bet the same bug type exists in the object decode path (where `{"a","b"}` is array-like illegal JSON) as it uses self::decode() recursion too.

GerHobbelt:
(Aside: Note that the CJSON::decode() is still not so sophisticated that it checks the legality of the Unicode characters in the Unicode strings in the input: some illegal stuff _may_ slip through that way. For when you feel like being extra nasty...)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -56,6 +56,7 @@ Version 1.1.16 under development
 - Bug #3406: Fixed Object of class Imagick could not be converted to string (eXprojects)
 - Bug #3410: Fixed wrong translation parameter name in CPhpAuthManager::addItemChild() (klimov-paul)
 - Bug #3569: CConsoleCommand::copyFiles() did not replace DIRECTORY_SEPARATOR correctly (jmxhit)
+- Bug #3589: CJSON::decode does not properly flag invalid JSON input as invalid (GerHobbelt)
 - Bug: Fixed the bug that backslashes are not escaped by CDbCommandBuilder::buildSearchCondition() (qiangxue)
 - Bug: Fixed URL parsing so it's now properly giving 404 for URLs like "http://example.com//////site/about/////" (samdark)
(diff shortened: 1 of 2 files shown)


samdark:
Won't be fixed in 1.1.

PR #3607, opened 2014-10-27
fix incorrect count of line breaks between client and server side

falexandrou:
Hello,

stumbled upon a bug where the line breaks are miscounted in client side.

with this bug the client side validation passes and server side validation fails

Please review


samdark:
Won't be fixed in 1.1.

Issue #3555, opened 2014-09-27
The method CActiveRecord::getOldPrimaryKey() returns a wrong value. 


iAchilles:
The method CActiveRecord::update() allows to specify list of attributes that must be saved. So, if the primary key attribute is manually assigned with a different value and list of attributes does not contain the name of the primary key, the primary key value will not be updated, but the method CActiveRecord::getOldPrimaryKey() will return the value that was manually set.

``` php
$model = Foo::model()->findByPk(1);
var_dump($model->getOldPrimaryKey()); // string(1) 1
$model->attr = 'abc';
$model->pk = 777;
$model->update(array('attr'));
var_dump($model->getOldPrimaryKey()); // int(777) 
```

ilyar:
:+1:

PR #3631, opened 2014-11-13
Fix for #3555

georaldc:
Fix for #3555 

CActiveRecord::_pk should only be updated if the primary key was included
as part of the attributes passed to CActiveRecord::update().

This should also fix the value returned by CActiveRecord::getOldPrimaryKey() when using the update method, without passing the primary key.


samdark:
Won't be fixed in 1.1.

PR #3691, opened 2015-01-02
added with criteria merging in CActiveRelation

### LINKED ISSUE #3579: [BUG] Criteria applied to CActiveRecord::getRelated does not merge correctly
I have traced it back to [CActiveFinder.php#L266](https://github.com/yiisoft/yii/blob/master/framework/db/ar/CActiveFinder.php#L266), but I am unable to solve this bug. 

Simple example below:

Comment.php

``` php
public function testScope()
{
    return $this->with(array(
        'user'=>array(
            'select'=>array('username'),
        ),
    ));
}
```

Controller.php

``` php
Comment::model()->findAll(array(
    'with'=>array(
        'user'=>array(
            'select'=>array('id'),
        ),
    ),
    'scopes'=>array('testScope'),
));
```

Above works as expected selecting both "id" and "username" from Comment relation "user".

``` php
$parentModel->getRelated('comments', true, array(
    'with'=>array(
        'user'=>array(
            'select'=>array('id'),
        ),
    ),
    'scopes'=>array('testScope'),
));
```

This should work too, but it doesn't, it only selects user's "id" because "testScope" is overwritten by the "with" declaration. 

Thank you for your feedback!

claudiocarotenuto:
fixes #3579

### DIFF
--- a/framework/db/ar/CActiveRecord.php
+++ b/framework/db/ar/CActiveRecord.php
@@ -2210,36 +2210,39 @@ class CActiveRelation extends CBaseActiveRelation
 				$this->on="({$this->on}) AND ({$criteria['on']})";
 		}
 
-		if(empty($this->with))
-			$this->with=$criteria['with'];
-		elseif(isset($criteria['with']))
+		if(isset($criteria['with']))
 		{
-			foreach($criteria['with'] as $k=>$v)
-			{
-				if(is_integer($k))
-					$this->with[]=$v;
-				elseif(isset($this->with[$k]))
-				{
-					$excludes=array();
-					foreach(array('joinType','on') as $opt)
-					{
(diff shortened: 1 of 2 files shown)


framework/db/ar/CActiveRecord.php
```diff
@@ -2210,8 +2210,37 @@ public function mergeWith($criteria,$fromScope=false)
 				$this->on="({$this->on}) AND ({$criteria['on']})";
 		}
 
-		if(isset($criteria['with']))
+		if(empty($this->with))
 			$this->with=$criteria['with'];
```
klimov-paul: This line will produce E_NOTICE in case `$criteria['with']` is not set.

samdark:
Won't be fixed in 1.1.

Issue #268, opened 2012-02-15
CDCriteria Count


qiangxue:
<b>What steps will reproduce the problem?</b>
1. $criteria-&gt;select = 'country.name'

<b>What is the expected output? What do you see instead?</b>

the total number of data record is wrong
when i try to getTotalItemCount();

<b>What version of the product are you using? On what operating system?</b>
1.1.8

<b>Please provide any additional information below.</b>

as the string 'country.name' starts with &quot;count&quot;,

Source code in CActiveFinder.php (Line: 742)

if($select!=='*' &amp;&amp; !strncasecmp($select,'count',5))
    $query-&gt;selects=array($select);

<p>Migrated from http://code.google.com/p/yii/issues/detail?id=3094</p><hr/><h2>earlier comments</h2>
<p><strong>mdomba said, at 2012-01-11T07:39:06.000Z:</strong></p>

<p>> as the string 'country.name' starts with "count"

did you mean count.name ?</p><p><strong>soul.geoffrey said, at 2012-01-11T22:22:19.000Z:</strong></p>

<p>i have a table called country
relation:
'applicant'=>array(self::BELONGS_TO,'Applicants','applicant_id'),
in the search() function
$criteria->with('applicant.country','applicant');
$criteria->search = 'country.name,......';
when i run
$dataProvider->getTotalItemCount();
the count is wrong.</p><p><strong>frieps said, at 2012-01-12T12:39:08.000Z:</strong></p>

<p>Related to bug #2904?
Maybe fix it by setting the totalitemcount in the return part of the dataprovider:
return new CActiveDataProvider(get_class($this), array(
  'criteria' => $criteria,
  'sort'=>$sort,
  'totalItemCount'=>count($this->model()->findAll($criteria)),
));

According to #2904, will be fixed in 1.1.10.</p><p><strong>soul.geoffrey said, at 2012-01-12T22:26:45.000Z:</strong></p>

<p>$dataProvider->getTotalItemCount();
will return the first country name, not the total number of records.

$criteria->select = 'country.name,applicant.name';  ===> wrong total

fixed by putting 'country.name' after the 'applicant.name'
$criteria->select = 'applicant.name,country.name';  ===> right total</p>

PR #3697, opened 2015-01-05
Fixes #3696 with broken case insensitiveness for Active Record count expressions

### LINKED ISSUE #3696: "count()" expression in criteria's "select" is not case insensetive since 1.1.16
In `1.1.15` this code 

```php
$criteria = [
    'select' => 'COUNT(DISTINCT group_id) AS count',
    'with' => [
        'category' => [
            'select' => false,
            ...
        ],
    ],
    'together' => true,
];
return Item::model()->published()->count($criteria);
```
generated next SQL : 
```sql
SELECT COUNT(DISTINCT group_id) AS count FROM ...
```

Since `1.1.16` with having this fixed `Bug #268: Fixed Active Record count error when some field name starting from 'count' (nineinchnick)` it generates **not** relevant SQL : 
```sql
SELECT COUNT(DISTINCT "t"."id") FROM ...
```

xt99:
This issue was introduced with `1.1.16` and fixed `Bug #268: Fixed Active Record count error when some field name starting from 'count' (nineinchnick)`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 Version 1.1.17 work in progress
 -------------------------------
 
+- Bug #3696: Fixed broken case insensitiveness for Active Record count expressions introduced with fixed #268 (xt99)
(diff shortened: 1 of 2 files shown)


qiangxue:
Thanks!

PR #3710, opened 2015-01-14
Update CActiveFinder.php

kosjak1:
$this->_pkAlias may be null, so it causing an error when trying to iterate over null value.

### DIFF
--- a/framework/db/ar/CActiveFinder.php
+++ b/framework/db/ar/CActiveFinder.php
@@ -825,7 +825,7 @@ class CJoinElement
 			else	// no matching related objects
 				return null;
 		}
-		else // is_array, composite key
+		else if(is_array($this->_pkAlias)) // is_array, composite key
 		{
 			$pk=array();
 			foreach($this->_pkAlias as $name=>$alias)
--- a/framework/db/ar/CActiveFinder.php
+++ b/framework/db/ar/CActiveFinder.php
@@ -825,17 +825,20 @@ class CJoinElement
 			else	// no matching related objects
 				return null;
 		}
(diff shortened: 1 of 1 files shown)


cebe:
how can this be null?

kosjak1:
I've made active relation between models with self::HAS_ONE, both models
have variable called house_id, so, when I tried to access house model
through address model - I've got error in framework files, which said me
that given value for foreach() is not an array. So, I made additional check
in framework, to avoid this error and now everything works fine.

Here is relation from model Address to model RnpHouseView:

'rnphouse' => array(self::HAS_ONE, 'RnpHousesView',
array('house_id'=>'house_id')),


One thing is that I changed a bit my previous patch to look like this,
because variable $pk is used later:

else  // is_array, composite key
{
$pk=array();
*if(is_array($this->_pkAlias)) {*
foreach($this->_pkAlias as $name=>$alias)
{
if(isset($row[$alias]))
$pk[$name]=$row[$alias];
else // no matching related objects
return null;
}
*}*
$pk=serialize($pk);
}



2015-01-14 13:36 GMT+02:00 Carsten Brandt <notifications@github.com>:

cebe:
Do you have a primary key defined for both tables?

kosjak1:
I figured that second model links to a view, and it don't have primary key.
I understand that we can redefine primary key in a model. But logically
would be to ignore null if table/view dont't have a primary key, to not
confuse Yii newbies.
14 Янв 2015 г. 19:37 пользователь "Carsten Brandt" <notifications@github.com>
написал:

cebe:
ActiveRecord can not work properly without a PK, so it is required to define one.

PR #3716, opened 2015-01-16
Fixes #3715  legacyDecrypt() not compatible with 1.1.15 decrypt()

### LINKED ISSUE #3715: CSecurityManager::legacyDecrypt() is not BC with 1.1.15 decrypt()
I discovered this in relation to https://github.com/yiisoft/yii/issues/3694 , and originally posted this as a comment there, but this is actually a distinct issue, separate from missing legacyEncrypt() function:

In the new implementation of legacyDecrypt, it always calls md5 on the key:
`$derivedKey=$this->substr(md5($key),0,mcrypt_enc_get_key_size($module));`

In the old decrypt method, it only ran the md5 if the key was not explicitly provided:
```
$key=$this->substr($key===null ? md5($this->getEncryptionKey()) : $key,0,mcrypt_enc_get_key_size($module));
```

The md5 call should be moved up in the legacyDecrypt() method to line 567 so that it is only called when the key has not been provided. Then the $derivedKey should run the substr method without the md5 modification of the key.  This will make it consistent with the old decrypt method.

Currently I *cannot* upgrade to 1.1.16 because none of my encrypted data can be decrypted, because we have used explicitly generated keys, which 1.1.16 is trying to run through md5 now.

DanaLuther:
Fixes https://github.com/yiisoft/yii/issues/3715

The old implementation of decrypt() only ran the md5 hash on the key if one was not explicitly provided.

### DIFF
--- a/framework/base/CSecurityManager.php
+++ b/framework/base/CSecurityManager.php
@@ -566,6 +566,7 @@ class CSecurityManager extends CApplicationComponent
 			$key=Yii::app()->getGlobalState(self::STATE_ENCRYPTION_KEY);
 			if(!$key)
 				throw new CException(Yii::t('yii','No encryption key specified.'));
+			$key = md5($key);
 		}
 
 		if(extension_loaded('mcrypt'))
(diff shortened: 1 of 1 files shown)


samdark:
Thanks! Will check it.

samdark:
Merged with https://github.com/yiisoft/yii/commit/c255d39d302dc360ce81e98048fe6b11b7cc1783. Thanks!

PR #3721, opened 2015-01-21
CStatElement work with PDO::CASE_UPPER

borro:


samdark:
Won't be fixed in 1.1.

PR #3735, opened 2015-01-31
$this->authTimeout missing null check

jmxhit:


samdark:
Original is correct. PHP is able to do `+ null` without issues. It equals to `+ 0`.

jmxhit:
The problem is that the state should not be set, or the next time call CWebUser's init will always logout the user when $this->absoluteAuthTimeout!==null && !$this->getIsGuest() evaluates true.

PR #3739, opened 2015-02-05
Added the ability to make the Redis cache connection optional

wisc:
When Yii fails to connect to Redis, it may not need to crash into a 500. The caching may be optional, as it would speed things up, but would not be required for the app to run. In that case a thrown exception is a bit hash.

This update creates a setting that lets the user toggle the behavior, defaulting to what it does now.

### DIFF
--- a/framework/caching/CRedisCache.php
+++ b/framework/caching/CRedisCache.php
@@ -60,6 +60,10 @@ class CRedisCache extends CCache
 	 */
 	public $database=0;
 	/**
+	 * @var boolean denotes whether the connection is optional or not
+	 */
+	public $successful_connection_optional=false;
+	/**
 	 * @var float timeout to use for connection to redis. If not set the timeout set in php.ini will be used: ini_get("default_socket_timeout")
 	 */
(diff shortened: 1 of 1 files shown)


cebe:
You can do this in your application if you want to skip it but I think this is nothing we want to have in the framework itself. Thanks for suggesting though.

wisc:
I don't see how I can skip this in my application. As soon as I set the cache to Redis, Yii will use it internally. So before any of my own code runs, the cache is already accessed by Yii internals. Thus I get a 500 without being able to do anything about it.

cebe:
You can override connect() like the following:

```php
protected function connect()
{
    try {
        parent::connect();
    } catch (CException $e) {
	Yii::app()->setComponent('cache', new CDummyCache);
        Yii::log('redis failed to connect: ' . $e->getMessage(), CLogger::LEVEL_ERROR);
    }

}
```

wisc:
Ah yes. Not a very elegant solution, but I guess there's little choice in the matter.

PR #3745, opened 2015-02-09
Fix for delete operation in gridview with non-ajax mode

psrustik:
When using CGridView with ajaxUpdate=false delete operation does not work. 
Client performs delete operation by GET method due to errors in the yiigridview.js.

See jquery.yiigridview.js:372

``` js
$.each(options.data, function (name, value) {
    /* ^^ error ^^ */
     $form.append($('<input type="hidden" name="t" value="" />').attr('name', name).val(value));
});
$form.submit();
```

jQuery.each() require list(array) for first argument, but code below create string:

``` js
/*  jquery.yiigridview.js:327*/
if (options.data === undefined) {
    options.data = $(settings.filterSelector).serialize();
}
```

This merge request fixed this problem and options.data converted to list (array).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,7 +3,7 @@
 
 Version 1.1.17 work in progress
 -------------------------------
-
+- Bug #3745: Fixed delete operation in gridview with non-ajax mode (psrustik)
 - Bug #2921: Fixed CStatePersister read/write concurrency issue causing state data corruption (matteosistisette, samdark)
 - Bug #3497: CErrorHandler messages for HTTP response codes were not matching RFCs (TeMPOraL)
 - Bug #3637: Fixed not quoting primary key in count statements (applee)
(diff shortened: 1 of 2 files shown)


miramir:
:+1:

thaddeusmt:
This seems to be the same error I ran in to, on the 1.1.16 branch, when ajaxUpdate is set to false on the CGridView.

`options.data` is a serialized string on line 363 where `$.each()` iterates over it. Prior to jQuery 1.9 `$.each()` just iterated over each character of the string, and nothing happened (no error either). Now, with jQuery 1.11, it throws this same error:

`Uncaught TypeError: Cannot use 'in' operator to search for`

It seems like it makes more sense to fix this by not serializing in the first place, on line `327` of the 1.1.16 branch:

```
options.data = $(settings.filterSelector).serialize();
```

to this:

```
options.data = $(settings.filterSelector);
```

The only possible issue I can see is that the `beforeAjaxUpdate()` callback might be expecting `options.data` to be serialized, and now it won't be.

samdark:
Won't be fixed in 1.1.

PR #3750, opened 2015-02-11
Fix empty array comparison in ORM

chenasraf:
When giving an empty array we should compare it like `addInCondition()` does, and not simply skip it. This might get rid of the whole `WHERE` clause completely and run updates on an entire table. `addInCondition()` already handles arrays with a low count so this will simply propagate the check to that method.


samdark:
Won't be fixed in 1.1.

Issue #3045, opened 2013-11-18
CGridView should have an option to only select/unselect rows by clicking on checkboxcolumn checkbox rather than anywhere in row


teo1978:
In a CGridView with selectableRows>1, by clicking anywhere on a row, you select it.
While this is wonderful to have this behavior as an option, and even OK to have it as a default, it's terribly annoying that you cannot disable it.

There should exist an option to disable it, so that only by clicking on a checkbox belonging to a checkboxcolumn you would select a row. 

It should be a property of CGridView. When true (default), rows would be selected by clicking anywhere within a row; when false, only clicking on checkboxcolumn checkboxes would select rows

Sibilino:
Is there still interest for this feature? It appears that the row selection function triggers from *tbody > tr* but that could be made configurable.

Why only on selectableRows > 1 though? Even with a single selectable row, it can be anoying to click on a row in order to copy some content and then have the row selected automatically.

Anyways, I already have some quick code to fix this issue on my app. Should I take the time to clean it up and make a pull request?

samdark:
Yes, issue is still open so a pull request would be appropriate.

PR #3762, opened 2015-02-20
#3045: CGridView: added rowSelectionSelector option to filter the elements in a row that can trigger a row selection on click

Sibilino:
Enh #3045 

Originally, when CGridView is configured to enable selectable rows, clicking anywhere on a row will select the row. However, the user may want to click on a row element without wanting to select the row; for example, when copying some row content to the clipboard.

After merge, CGridView will accept the `rowSelectionSelector` option, which specifies a jQuery selector so that only the row's children that are matched will trigger the row's selection. The default value for this option is an empty string, which leaves the original selection behavior unmodified.

For example:

``` php
//...
'rowSelectionSelector' => 'input.select-on-check',
//...
```

Will restrict row selection to clicking on the checkboxes generated by a CCheckBoxColumn.


samdark:
Won't be merged in 1.1.

Issue #3279, opened 2014-03-19
cjuiautocomplete - use (active)SearchField


Leffe108:
A common use case for auto complete is to use on a search field. However, it is currently not possible to make CJuiAutoComplete use (active)SearchField instead of (active)TextField other than sub classing and overriding the whole run() method (which is almost the entire widget code).

Here is four arguments for using the search field for search use cases:
1. It is more semantically correct
2. Screen readers will announce it as a search field instead as a text field
3. Many AT (assistive technology) have shortcuts to jump to search fields.
4. In iOS and possible also Android, the virtual screen keyboard will show a Search button instead of the Okay button.

Thus I propose that it should be possible to configure CJuiAutoComplete so that a search field will be generated instead of a text field. I guess there are genuine use cases when text field is preferred so a commit need to either add an configuration option or move rendering of the input field to a protected method that can be overriding without having to duplicate the whole run() code in your application.

samdark:
Won't be fixed in 1.1.

PR #3765, opened 2015-02-20
Enh #3279: CJuiAutoComplete: added `inputType` option that controls the type of the generated input field

Sibilino:
Enh #3279

Originally, the CJuiAutoComplete widget always generated an input of type="text". The only way of changing the input type was to override the run() method, which contained most of the widget's code.

After merge, the CJuiAutoComplete widget will accept an additional option `inputType` which controls the autocomplete input's type. Its default value is `'text'` and accepts either `'text'`, `'search'`, `'email'`, `'tel'` or `'url'`.

Furthermore, the HTML-generating code has been moved to a separate protected function for easier overriding.


samdark:
Won't be fixed in 1.1.

PR #3767, opened 2015-02-23
generateRandomDigits added to CSecurityManager

fazliddin:
Sometimes we need random string only consisting from digits. It may be very handy, for example, for generating password recovery token which is sent by SMS.

### DIFF
--- a/framework/base/CSecurityManager.php
+++ b/framework/base/CSecurityManager.php
@@ -353,6 +353,28 @@ class CSecurityManager extends CApplicationComponent
 		$key=$this->substr($key,0,64);
 		return $func((str_repeat(chr(0x5C), 64) ^ $key) . pack($pack, $func((str_repeat(chr(0x36), 64) ^ $key) . $data)));
 	}
+	
+	/**
+	 * Same as {@link generateRandomString} with one exception that it generates only [0-9] digits.
+	 * @param integer $length length of the generated string in characters.
+	 * @param boolean $cryptographicallyStrong set this to require cryptographically strong randomness.
+	 * @return string|boolean random string or false in case it cannot be generated.
+	 * @since 1.1.17
+	 */
+	public function generateRandomDigits($length, $cryptographicallyStrong=true)
+	{
+		$str = $this->generateRandomString($length, $cryptographicallyStrong);
+        	if($str === false)
+        		return false;
+        		
+        	$result = '';
+        	for($i=0; $i<strlen($str); $i++)
+        	{
+        		$result .= ord($str{$i}) % 10;
+        	}
+        
+        	return $result;
+	}
 
 	/**
 	 * Generate a random ASCII string. Generates only [0-9a-zA-z_~] characters which are all


samdark:
I don't think it should be in the core:

1. Generation algorithm doesn't look secure. It looses precision.
2. It's application specific.

fazliddin:
Ok, no problem. Thanks for note about insecurity. Since I am using this method on my project, could you explain what do you mean by precision. Does cryptographically strong digits mean unpredictable ones?

samdark:
Yes, it means unpredictable ones. What I don't like is:

```php
$result .= ord($str{$i}) % 10;
```

ord for a = 97, ord for k = 107. In both cases you're getting `7` after doing `%`. So overall you're getting about 4 times less unique values and I'm not sure about it.

fazliddin:
Char from [A-Za-z0-9~\_] set may have 62 different values (25+25+10+2), on the other hand char from [0-9] set has only 10 different values. It means string from only digits has 6.2 times lesser unique values than that of [A-Za-z0-9~\_]. And there is no way to improve this. You are right that digit of a and k match, but even if you change algorithm so that digit of k were different, then one can always find some other 5 variants from [A-Za-z0-9~\_] set which matches with digit of a. I compared my method with following:
```
$result = ' ';
        for($i=0; $i<$length; $i++)
            $result .= mt_rand(0,9);
```
And from result one can not say that one of them is better
```
ord           mt_rand
---------     ----------
355152     493558
921163     493758
091637     065955
799656     133674
720017     203234
786102     172837
053649     068770
182099     679965
156823     815887
922825     385455
199523     587408
353716     014853
253326     871477
769007     056037
756293     030609
588877     692975
706721     011163
018750     871963
900975     906099
586932     411237
755219     627521
706042     799046
853404     985636
328053     875214
049808     425407
807976     370889
917961     325919
988668     907614
557967     190768
109120     672105
299869     215408
761528     138304
939197     769043
036994     740839
498746     858917
367659     823466
750460     590185
189200     721160
304868     720301
086617     123906
951551     617807
951058     896093
679903     357189
572786     231712
820053     519880
391290     494831
460504     796963
721689     951222
407146     436068
339699     639288
082295     115537
175859     158471
781431     198567
655319     584951
679197     187605
099885     322553
639598     512643
050916     564752
279725     639198
808691     115941
806590     323085
622559     899170
765291     530987
729383     151608
169226     709144
911594     348318
190420     406027
668184     684662
488020     012340
906060     896569
697275     017659
390016     216219
850527     865747
519841     079421
122525     184308
260607     460703
743951     256673
495956     260776
050764     371653
593538     434334
306142     212846
150545     898683
073276     797122
648528     027715
147211     744410
769829     413833
819031     145658
853968     678358
459118     240923
673918     488732
218934     188886
763751     973234
984185     192163
764547     865773
320922     664196
895937     759305
586932     042604
085061     490377
237047     349452

samdark:
Well, probably it's OK for the use case. Still I think it's a bit too specific to be in the core framework.

PR #3768, opened 2015-02-23
Update pl_pl.php

winiarekk:
Fixed currencyFormat for negative values


cebe:
this data is imported from the unicode CLDR project, we do not want to change it ourselfs. Please report problems to them.

Issue #3773, opened 2015-02-25
CFileHelper crashes on broken symlinks


etienneq:
This applies to findFiles() and copyFiles() as well. The reason is in the implementation of findFilesRecursive() and copyDirectoryRecursive()

As is_file() returns false an broken symlinks the code asumes that these are directories. To fix this it should explicitly tested with is_dir() before the recursive call takes place.

I will provide a pull request shortly.

Cheers
Etienne

samdark:
Won't work on Windows. There non-broken symlinks return is_dir true.

samdark:
How to get such broken link?

etienneq:
Even when the symlink points to a regular file? It tested in on Linux.
There is_file() and is_dir() work fine on symlinks.

2015-02-25 10:42 GMT+01:00 Alexander Makarov <notifications@github.com>:

etienneq:
@samdark First create a symlink, then delete the original file or directory.

samdark:
Won't work on Windows where `is_dir` still returns true even if original dir was deleted.

PR #3774, opened 2015-02-25
#3773: CFileHelper does no longer crash on broken symlinks

etienneq:


samdark:
#3773

samdark:
Won't be fixed in 1.1.

Issue #3457, opened 2014-07-18
Error with PUT parameters in CHttpRequest ( method getRestParams() )


pedrocalgaro:
Hi everyone,

I'm having a problem when I trying to get parameters from a PUT request and I believe is is a bug of the framework.
When the method getRestParams() is called, it tries to parse the rawBody.
The method getRawBody() returns the value of file_get_contents('php://input')

The function file_get_contents('php://input') returns the value:

```json
{"foo": "foo content", "bar": "bar content"}
```

The function getRestParams() tries to parse this content calling the function mb_parse_str

```php
mb_parse_str($this->getRawBody(), $result);
```

But this returns:
```php
Array
(
    [{"foo": "foo content", "bar": "bar content"}] => 
)
```
So I changed the line: (https://github.com/yiisoft/yii/blob/master/framework/web/CHttpRequest.php#L290)

```php
mb_parse_str($this->getRawBody(), $result);
```

To:
```php
$result = json_decode($this->getRawBody(), true);
```

And after that, it works like a charm.

You could look at it?

PR #3780, opened 2015-02-27
Enh #3457: CHttpRequest able to decode JSON REST parameters.

Sibilino:
Enh #3457

Originally, REST requests such as PUT, PATCH and DELETE  had their body parsed using `mb_parse_str()` or `parse_str()`, which means that JSON data would not be properly parsed.

The merge is basically a backport of the solution existing in Yii 2. It will add `CJSON::decode()` to the possible parsing functions, activated only if the request's content type is `application/json`. This content type can be retrieved by the new public function `getContentType()`.

In addition, as in Yii 2, the JSON decoding function can be configured to use associative arrays (by default) for JSON object data, according to the new boolean property `$jsonAsArray`.

### DIFF
--- a/framework/web/CHttpRequest.php
+++ b/framework/web/CHttpRequest.php
@@ -293,7 +293,7 @@ class CHttpRequest extends CApplicationComponent
 		if($this->_restParams===null)
 		{
 			$result=array();
-			if ($this->getContentType() === 'application/json')
+			if (strncmp($this->getContentType(), 'application/json', 16) === 0)
 				$result = CJSON::decode($this->getRawBody(), $this->jsonAsArray);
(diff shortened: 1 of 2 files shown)


framework/web/CHttpRequest.php
```diff
@@ -287,7 +293,9 @@ public function getRestParams()
 		if($this->_restParams===null)
 		{
 			$result=array();
-			if(function_exists('mb_parse_str'))
+			if ($this->getContentType() === 'application/json')
```
cebe: this should check only the beginning of content type to also support json which has the addition `; charset=utf8` like this: `strncmp($this->getContentType(), 'application/json', 16) === 0`

PR #3786, opened 2015-03-04
namespace may be absent in node of method name

Alxdhere:


samdark:
Won't be fixed in 1.1.

PR #3787, opened 2015-03-04
Wrong pager url in GridView when ajaxUrl attribute is set

odya:
Wrong pager url in GridView when ajaxUrl attribute is set


samdark:
Won't be fixed in 1.1.

PR #3796, opened 2015-03-12
additional soap directives

Alxdhere:
@soap-dependent - add specified class in wsdl
@soap-extra - add specified xml in wsdl


samdark:
Won't be fixed in 1.1.

PR #3804, opened 2015-03-14
Fixed CPgsqlColumnSchema.php to support resource type.

sannek8552:
Fixed CPgsqlColumnSchema.php to support resource type.

It is related to yiisoft#1181

When we load model with bytea column dbtype - Yii column type is "string" by default. (type of data is resource)
In this case when we trying to save this model after execution method \CDbColumnSchema::typecast value cast to string. And we have Resource #2 string in database.


samdark:
Won't be fixed in 1.1.

Issue #3812, opened 2015-03-21
Add validateValue method to CBooleanValidator class


ua2004:
CEmailValidator class has a convenient method validateValue if we simply want to validate an email without using a model. It would be great to add the same method for validating boolean values.

Issue #3764, opened 2015-02-20
incorrect doc for CEmailValidator::validateValue()


fazliddin:
It is written that
```
Note that this method does not respect {@link allowEmpty} property.
```
But look at the code
```
if($this->allowEmpty && $this->isEmpty($value))
			return true;
...
```

cebe:
introduced by #3283, should be reverted.

Lexxtor:
```
Note that this method does not respect {@link allowEmpty} property.
```
Why it was done like that?

cebe:
because `validateValue()` is to be used to check if a string is a valid email without the context of the model validation flow. It should be possible to just pass a string and check whether it is a valid email without having to check if it is empty or not before.

PR #3813, opened 2015-03-21
Added 'validateValue' method to 'CBooleanValidator' class, enh #3812

ua2004:
#3812

### DIFF
--- a/framework/validators/CBooleanValidator.php
+++ b/framework/validators/CBooleanValidator.php
@@ -66,6 +66,25 @@ class CBooleanValidator extends CValidator
 			));
 		}
 	}
+	
+	/**
+	 * Validates a static value to see if it is a valid boolean.
+	 * This method is provided so that you can call it directly without going
+	 * through the model validation rule mechanism.
+	 * @param mixed $value the value to be validated
+	 * @return boolean whether the value is a valid boolean
+	 * @since 1.1.17
+	 */
+	public function validateValue($value)
+	{
+		if($this->allowEmpty && $this->isEmpty($value))
+			return true;
+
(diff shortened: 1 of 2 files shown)


framework/validators/CBooleanValidator.php
```diff
@@ -66,6 +66,25 @@ protected function validateAttribute($object,$attribute)
...
+	 * Validates a static value to see if it is a valid boolean.
+	 * This method is provided so that you can call it directly without going
+	 * through the model validation rule mechanism.
+	 * @param mixed $value the value to be validated
+	 * @return boolean whether the value is a valid boolean
+	 * @since 1.1.17
+	 */
+	public function validateValue($value)
+	{
+		if($this->allowEmpty && $this->isEmpty($value))
```
cebe: this check should not be done here, and the method could be used in validateAttribute then.

see also #3764

ua2004: @cebe 'validateAttribute' now uses 'validateValue' method. However, why shouldn't this check be done here? The same check is used in 'validateValue' method in CEmailValidator class.

cebe: it was accidentally introduced in email validator which broke BC and does not make sense.

ua2004: But then where is 'allowEmpty' parameter considered?

cebe: only in validateAttribute

samdark:
@cebe is it ready to be merged?

cebe:
not sure, need to check it.

cebe:
Merged, with adjustments, thanks!

Issue #3819, opened 2015-03-25
Error in framework unit tests


zdrsoft:
YII Version 1.1.16
Ubuntu 14.10
PHP Version 5.5.12-2ubuntu4.3

When I run unit tests of the frame work. The tests stop and returns an error:

# /yii/tests$ phpunit framework/

PHPUnit 4.5.0 by Sebastian Bergmann and contributors.

Configuration read from /var/www/html/zzz/yii/tests/phpunit.xml

........................   61 / 1172 (  5%)
........................  122 / 1172 ( 10%)
........................  183 / 1172 ( 15%)
........................  244 / 1172 ( 20%)
........................  305 / 1172 ( 26%)
........................  366 / 1172 ( 31%)
........................  427 / 1172 ( 36%)
........................  488 / 1172 ( 41%)
........................  549 / 1172 ( 46%)
........................  610 / 1172 ( 52%)
........................  671 / 1172 ( 57%)
........................ PHP Fatal error:  Call to a member function createCommand() on a non-object in /var/www/html/zzz/yii/tests/framework/validators/CExistValidatorTest.php on line 37

cebe:
they are running fine on travis: https://travis-ci.org/yiisoft/yii/jobs/55159304
have you configured databases correctly?

zdrsoft:
I found how could be reproduced the error. If you do not have installed "php5-sqlite" extension and run the tests the error will rise, even in the code is placed check for existing php sqllite installation in the same class take a look at "setUp" methods of this classes

yii/tests/framework/validators/CExistValidatorTest.php
yii/tests/framework/validators/CUniqueValidatorTest.php

I have provided a bug fix.  	
Fix foor #3819

PR #3820, opened 2015-03-25
Fix for #3819

zdrsoft:
I found how could be reproduced the error. If you do not have installed "php5-sqlite" extension and run the tests the error will rise, even in the code is placed check for existing php sqllite installation in the same class take a look at "setUp" method.

Bug fix applied:
Affected files: 
	tests/framework/validators/CExistValidatorTest.php
	tests/framework/validators/CUniqueValidatorTest.php
4a7127b..16410a6  master -> master

Fix for #3819


cebe:
Good find, thanks!

PR #3827, opened 2015-03-26
CRedisCache Socket Options

yoshida-mediba:
When you connect to every access to the Redis server application is slow.
Therefore I think it is often used STREAM_CLIENT_PERSISTENT option.


cebe:
can you add a changelog line?

yoshida-mediba:
done

cebe:
Merged, thanks!

yoshida-mediba:
:+1:

PR #3847, opened 2015-04-16
Fix #3846: Fetch as CDbDataReader for cases expecting arrays

### LINKED ISSUE #3846: Fetch as CDbDataReader for cases expecting arrays
Otherwise it causes fatal error of memory exhausting for big queries as it deals with arrays of incredible size.

PatchRanger:
Otherwise it causes fatal error of memory exhausting for big queries as it deals with arrays of incredible size.
Fix #3846.

PS Sorry for alien commit, it is from another pull-request https://github.com/yiisoft/yii/pull/3788 .

### DIFF
--- a/framework/db/CDbCommand.php
+++ b/framework/db/CDbCommand.php
@@ -509,14 +509,18 @@ class CDbCommand extends CComponent
 			else
 				$this->_statement->execute($params);
 
-			if($method==='')
-				$result=new CDbDataReader($this);
-			else
-			{
+			if (in_array($method,['fetchColumn','fetch'])) {
 				$mode=(array)$mode;
 				call_user_func_array(array($this->_statement, 'setFetchMode'), $mode);
 				$result=$this->_statement->$method();
 				$this->_statement->closeCursor();
+			}
+			else {
+				$result = new CDbDataReader($this);
+				if (!empty($mode)) {
+					// $mode could be an array - so can't use just $result->setFetchMode($mode).
+					call_user_func_array([$result,'setFetchMode'],(array)$mode);
+				}
 			}
 
 			if($this->_connection->enableProfiling)
--- a/framework/web/CController.php
+++ b/framework/web/CController.php
@@ -390,9 +390,10 @@ class CController extends CBaseController
 	protected function replaceDynamicOutput($matches)
 	{
 		$content=$matches[0];
-		if(isset($this->_dynamicOutput[$matches[1]]))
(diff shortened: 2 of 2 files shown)


framework/db/CDbCommand.php
```diff
@@ -509,15 +509,19 @@ private function queryInternal($method,$mode,$params=array())
...
 				$mode=(array)$mode;
 				call_user_func_array(array($this->_statement, 'setFetchMode'), $mode);
 				$result=$this->_statement->$method();
 				$this->_statement->closeCursor();
 			}
+			else {
+				$result = new CDbDataReader($this);
+				if (!empty($mode)) {
+					// $mode could be an array - so can't use just $result->setFetchMode($mode).
+					call_user_func_array([$result,'setFetchMode'],(array)$mode);
```
klimov-paul: This make no sense: why `CDbDataReader::setFetchMode()` should be invoked here?

framework/web/CController.php
```diff
@@ -390,9 +390,10 @@ public function processDynamicOutput($output)
 	protected function replaceDynamicOutput($matches)
 	{
 		$content=$matches[0];
-		if(isset($this->_dynamicOutput[$matches[1]]))
+		if(isset($this->_dynamicOutput[$matches[1]]) && is_array($this->_dynamicOutput[$matches[1]]))
```
klimov-paul: What is this change?
How does it related to #3846?

PatchRanger: > How does it related to #3846?

As I've mentioned before, that is from another pending pull-request #3788. It is not related to this issue, it is here just because I did not find another way to contribute having pending pull-request at `master` branch. I guess the pull-request should be re-rolled - but at the time I don't see a chance. Please re-roll yourself as my `master` branch already has that commit and I can't create a fork from scratch.

framework/db/CDbCommand.php
PatchRanger: Because in any case of using the result of fetching (either by `read()` and `readAll()` or as implementor of `Iterator` interface), it depends on such constructions as `$this->_statement->fetch()` or `$this->_statement->fetchAll()` which in turn depend of fetch mode - that is why we should take that option into account. Different values of that option mean different final result of the fetching - we can't just ignore it.

klimov-paul: Why don't use:
```
$reader = $command->query();
$reader->setFetchMode();
```
?

PatchRanger: I've answered there: https://github.com/yiisoft/yii/issues/3846#issuecomment-93947920 .

PR #3849, opened 2015-04-22
Fixed #2384: Render dynamicContent at page render, not at the moment of ...

### LINKED ISSUE #2384: CClientScript ignores DynamicCaching
``` php
...other HTML content...
<?php if($this->beginCache($id)) { ?>
...fragment content to be cached...
    <?php $this->renderDynamic($callback); ?>
...fragment content to be cached...
<?php $this->endCache(); } ?>
...other HTML content...
```

In any of the register*()-functions of CClientScript, a caching action is recorded.
This is (incorrectly) also happening when renderDynamic() is handling it's callback.

**Expected**
Any calls to CClientScript from renderDynamic($callback) should not be cached and replayed.

PatchRanger:
...placing into cache

### DIFF
--- a/framework/web/CController.php
+++ b/framework/web/CController.php
@@ -390,9 +390,10 @@ class CController extends CBaseController
 	protected function replaceDynamicOutput($matches)
 	{
 		$content=$matches[0];
-		if(isset($this->_dynamicOutput[$matches[1]]))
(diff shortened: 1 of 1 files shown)


samdark:
#2384

samdark:
Won't be fixed in 1.1.

PR #3850, opened 2015-04-22
Fix #1588: Added aliases for 'through' tables

### LINKED ISSUE #1588: Not unique table/alias mySql exception with two through relations on another through relation
Let's say we have Author, Category, AuthorHasCategory. Each author has a main category he is working on. Therefore the table "AuthorHasCategory" has a boolean "is_main". Category has n:1 relations to Foo and Bar to further describe the Category.

Now we define following relations on Author:

```
public function relations() {
    return array(
            'authorHasCategories' => array(self::HAS_MANY, 'AuthorHasCategory', 'author_id'),
            'mainCategory'=>array(self::HAS_ONE,'Category',array('category_id'=>'id'),'through'=>'authorHasCategories', 'on' => 'is_main = 1'),
            'mainCategoryFoo' => array(self::HAS_ONE, 'Foo', array('foo_id' => 'id'), 'through' => 'mainCategory'),
            //BUG: adding following declaration will not work due to a duplicate table alias exception from MySql
            'mainCategoryBar' => array(self::HAS_ONE, 'Bar', array('bar_id' => 'id'), 'through' => 'mainCategory')
    );
}
```

When mainCategoryFoo and mainCategoryBar are **added to "with" at the same time**, an mysql exception occurs:

```
SQLSTATE[42000]: Syntax error or access violation: 1066 Not unique table/alias: 'authorHasCategories'.
```

When checking the sql query you will find twice:

```
 LEFT OUTER JOIN `author_has_category` `authorHasCategories` ON (`authorHasCategories`.`author_id`=`t`.`id`)
```

The first time for the actual pivot table and another time it seems the whole left other join chain is added another time instead of reuse existing joins.

Sorry for the abstract example, I had to cut it out from our current project.

PatchRanger:
Extended version of commit, provided by @acorncom.

### DIFF
--- a/framework/db/ar/CActiveFinder.php
+++ b/framework/db/ar/CActiveFinder.php
@@ -219,8 +219,21 @@ class CActiveFinder extends CComponent
 				$with=substr($with,0,$pos);
 			}
 
+			$childAlias = isset($options['alias']) ? $with.'_'.$options['alias'] : $with;
 			if(isset($parent->children[$with]) && $parent->children[$with]->master===null)
-				return $parent->children[$with];
+			{
+				if($childAlias != $with && !isset($parent->children[$childAlias]))
+				{
+					if(!isset($parent->children[$with.'_'.$options['alias']]))
+					{
+						$child = clone $parent->children[$with];
+						$child->tableAlias = $options['alias'];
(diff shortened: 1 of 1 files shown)


samdark:
Won't be fixed in 1.1.

PR #3853, opened 2015-04-23
Allow dotted notation while specifying index

### LINKED ISSUE #3852: CActiveFinder: allow dotted notation while specifying index
It would be great if we can specify "index" by using dotted notation for relational references. For example:

```
public function relations() {
    return array_merge(parent::relations(),array(
        'identities' => array(self::HAS_MANY, 'UserIdentity', 'user_idx','index'=>'type.id'),
    ));
}
```

I think It is possible as it seems "indexing" in Yii happens after relations are fetched.

ab-kily:
It resolves #3852

### DIFF
--- a/framework/db/ar/CActiveFinder.php
+++ b/framework/db/ar/CActiveFinder.php
@@ -519,7 +519,7 @@ class CJoinElement
 			foreach($child->records as $record)
 			{
 				if($child->relation->index!==null)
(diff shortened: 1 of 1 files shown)


cebe:
this change is incomplete, there are more places where index is relevant. I am also not sure if we should introduce this because a relation may not exist which will cause the indexing to fail.

Issue #3549, opened 2014-09-23
Option to Disable CGridView Filter Submission on Tab Keypress


iwanjunaid:
Isn't it would be nice if there is such an option for disabling CGridView filter submission on tab keypress on  filter cell ?

moshe770:
hi guys i am interested in this issue of allowing filter on enter key press only. how did you solve it?

PR #3855, opened 2015-04-24
Enhancement for #3549

georaldc:
- Prevent text element filters from being triggered on change event. Text
  filters will only fire on ENTER keypress. Other elements like Select
  dropdowns' functionality should stay unaffected.

This should stop tab keypresses in cgridview from triggering filters

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 Version 1.1.17 work in progress
 -------------------------------
 
+- Enh #3549: Prevent text filters from being triggered on change event. Only trigger when using ENTER key (georaldc)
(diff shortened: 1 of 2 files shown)


georaldc:
I never really had a problem with cgridview's filters but I have seen instances where users would type a search value, hit tab to type another search value on another column and suddenly have the cgridview refresh and clear out whatever they were typing in the 2nd search box due to the filter getting triggered for the 1st search value. This fix *hopefully* remedies that since users tend to hit Enter to trigger their search.

samdark:
Closing since 1.1 isn't accepting enhancement anymore.

PR #3859, opened 2015-05-03
Update CActiveRecord.php

sinner:
Add support to really schema caching duration


samdark:
Any explanation of what patch does and why it does so?

cebe:
schema caching is handled in db connection and not AR itself.

PR #3865, opened 2015-05-12
when use the reverse proxy ,the app will still get the real client ip.

### LINKED ISSUE #3864: how get real ip when using lvs?
it likes the [getUserHostAddress](https://github.com/yiisoft/yii/blob/master/framework/web/CHttpRequest.php#L724)  process the case,or have any other methods to process this?

noname007:
Fix #3864.when use the reverse proxy ,the app will still get the real client ip. Default  http header key  would  be  x-forwarded-for

### DIFF
--- a/framework/web/CHttpRequest.php
+++ b/framework/web/CHttpRequest.php
@@ -712,8 +712,15 @@ class CHttpRequest extends CApplicationComponent
 	 * Returns the user agent, null if not present.
 	 * @return string user agent, null if not present
 	 */
-	public function getUserAgent()
+	static function getUserHostAddress($proxy_set_header ='HTTP_X_FORWARDED_FOR')
 	{
+		if(isset($_SERVER[$proxy_set_header])){
+			$ip = $_SERVER[$proxy_set_header];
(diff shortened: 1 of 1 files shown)


framework/web/CHttpRequest.php
```diff
@@ -712,8 +712,15 @@ public function getUrlReferrer()
 	 * Returns the user agent, null if not present.
 	 * @return string user agent, null if not present
 	 */
-	public function getUserAgent()
-	{
+	static function getUserHostAddress($proxy_set_header ='HTTP_X_FORWARDED_FOR')
+	{
```
klimov-paul: This is definetely incorrect change: why `getUserAgent()` removed and `getUserHostAddress()` is static?

samdark:
We won't add it to the framework since it's a security issue unless you're 100% sure you're behind a proxy.

framework/web/CHttpRequest.php
noname007: Sorry !   copy wrong  because of my  mindless...

PR #3869, opened 2015-05-19
Bug fix in jQuery Yii GridView plugin

a-t:
When updating grid with a POST request, and the request data is not provided and thus is collected from form with the jQuery `.serialize()` method, the script fails at [this line](https://github.com/yiisoft/yii/blob/8c30239dce366a70bc7486cd605c00c4938f0c9e/framework/zii/widgets/assets/gridview/jquery.yiigridview.js#L366) because `options.data` is a string at this point and can't act as an argument for `$.each()`. This fix makes `options.data` a plain object instead of a string in the described circumstances.


a-t:
I've found that in Yii 2 the corresponding piece of code is already written in a similar to the provided in this merge request way. You can see it [here](https://github.com/yiisoft/yii2/blob/2.1/framework/assets/yii.gridView.js#L87).

samdark:
Please add a line for changelog and I'll merge it.

a-t:
Done. Is it OK?

samdark:
Yes. Thank you!

a-t:
You're welcome :)

PR #3871, opened 2015-05-22
Added module class example to module guide.

cn007b:
Hi.
Not sure that i done all correct... but i just try to add module class example.
Hope it will be helpful, and more obviously and transparent for beginners to create modules.

### DIFF
--- a/docs/guide/basics.module.txt
+++ b/docs/guide/basics.module.txt
@@ -69,6 +69,16 @@ $postPerPage=Yii::app()->controller->module->postPerPage;
 // $postPerPage=$this->module->postPerPage;
 ~~~
 
+Now you need create module class, if it not done yet.
+For example, the `forum` module should have in file `protected/modules/forum/ForumModule.php` class `ForumModule`:
+
+~~~
+[php]
+class ForumModule extends CWebModule
+{
+}
+~~~
+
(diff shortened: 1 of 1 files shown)


docs/guide/basics.module.txt
```diff
@@ -69,6 +69,16 @@ $postPerPage=Yii::app()->controller->module->postPerPage;
...
+Now you need create module class, if it not done yet.
+For example, the `forum` module should have in file `protected/modules/forum/ForumModule.php` class `ForumModule`:
+
+~~~
+[php]
+class ForumModule extends CWebModule
+{
+}
+~~~
+
```
cebe: isn't this already explained above under " Creating Module"?

codeName007: Yeah it's explained above, you are right!
But it little bit implicit. I decided to show example to facilitate understanding.
When i tried to create module first time long time ago, it was hard to understand without examples. And now i see that it's still hard to understand for other people, for example: [question 1](http://stackoverflow.com/questions/30279114/how-can-i-get-module-from-subfolder/30280011#30280011), [question 2](http://stackoverflow.com/questions/30253627/how-to-setup-second-module-in-yii/30254493#30254493)...
Yes my example looks ugly (despite that it's so simple and so thin) maybe need to place it to section "Creating Module" after mention about CWebModule? And extend with "init" method or some else?

samdark:
Won't be merged but thanks for suggestion.

PR #3878, opened 2015-06-01
Fix #3800 CPropertyValue::ensureArray documentation

### LINKED ISSUE #3800: Incorrect documentation for CPropertyValue::ensureArray()
Link: http://www.yiiframework.com/doc/api/1.1/CPropertyValue

1) Describtion in "Public Methods" section is not finished: "Converts a value to array type. If the value is a string and it is" (...and it is what?)
2) Detailed description is also incomplete: "If the value is a string and it is not in this form then an array consisting of just the string will be returned. If the value is not a string then" (...then what?)
http://www.yiiframework.com/doc/api/1.1/CPropertyValue#ensureArray-detail

fbau123:


samdark:
Thanks!

Issue #3438, opened 2014-07-08
[Postgresql] Wrong behaviour for multiple constraints for the same field


aztech:
I have defined following constraints:
1. FOREIGN KEY (field_1, field_2) REFERENCES table_a(ref_field_a, ref_field_b)
2. FOREIGN KEY (field_1) REFERENCES table_b(ref_field_c)
3. FOREIGN KEY (field_2) REFERENCES table_c(ref_field_d)

In method findForeignKey of class CPgsqlSchema variable $table->foreignKeys keeps information about foreign keys on field level but the logic of the methods allows to have just 1 constraint per 1 field only.

In my case $table->foreignKeys after processing 1st fk keeps following information
array(
  field_1 => array(table_a, ref_field_a),
  field_2 => array(table_b, ref_field_b)
)
After processing 2nd foreign key
array(
  field_1 => array(table_c, ref_field_c),
  field_2 => array(table_b, ref_field_b)
)
After processing 3rd key
array(
  field_1 => array(table_c, ref_field_c),
  field_2 => array(table_d, ref_field_d)
)

As you see the information about 1st foreign key is lost.

samdark:
Won't be fixed in 1.1.

PR #3882, opened 2015-06-06
Postgres - ignore composite foreign keys. 

baso10:
Fix for #3438
foreignKeys property should not contain invalid records.


samdark:
Won't be fixed in 1.1.

Issue #3890, opened 2015-06-16
CWebService call to HTTP_RAW_POST_DATA throws deprecated warning in PHP 5.6


marcovtwout:
In my error log:
`PHP Deprecated:  Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead.`

Probably caused by this line: https://github.com/yiisoft/yii/blob/master/framework/web/services/CWebService.php#L239

More Info: http://php.net/manual/en/reserved.variables.httprawpostdata.php

marcovtwout:
This could be a PHP 5.6 bug for the default value: http://stackoverflow.com/a/28291687/729324

PR #3891, opened 2015-06-16
#3890 - Always use `php://input` since HTTP_RAW_POST_DATA is deprecated

AloneCoder:
HTTP_RAW_POST_DATA is deprecated in since of PHP 5.6.0 and causes a warning

### DIFF
--- a/framework/web/services/CWebService.php
+++ b/framework/web/services/CWebService.php
@@ -236,10 +236,7 @@ class CWebService extends CComponent
 	{
 		if($this->_method===null)
 		{
-			if(isset($HTTP_RAW_POST_DATA))
-				$request=$HTTP_RAW_POST_DATA;
-			else
-				$request=file_get_contents('php://input');
+			$request=file_get_contents('php://input');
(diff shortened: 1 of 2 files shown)


framework/web/services/CWebService.php
```diff
@@ -236,10 +236,7 @@ public function getMethodName()
 	{
 		if($this->_method===null)
 		{
-			if(isset($HTTP_RAW_POST_DATA))
-				$request=$HTTP_RAW_POST_DATA;
-			else
-				$request=file_get_contents('php://input');
+			$request=file_get_contents('php://input');
```
samdark: `php://input` can be read only once.

lcacciagioni:
@samdark maybe you can try with something like this:
```php
if (version_compare(phpversion(), '5.6.0', '<') && isset($HTTP_RAW_POST_DATA)) {
    $request=$HTTP_RAW_POST_DATA;
} else {
    $request=file_get_contents('php://input');
}
```
I'm having the same issue and I'll try it this week, I'll keep you updated.

lcacciagioni:
BTW I have added the following to the php.ini to avoid the deprecation message in 5.6.
```
always_populate_raw_post_data=-1
```

PR #3895, opened 2015-06-20
CActiveRecord::model() defaults to called class

adeelnawaz:
*Purpose*
The `CActiveRecord::model()` method uses `__CLASS__` magical constant as a default fallback for class name. It must not use it as this constant is early static binding, hence it will always return *CActiveRecord*. Instead, the class must use PHP `get_called_class()` (PHP 5 >= 5.3) to get late static binding class name. So that every class that implements `CActiveRecord` doesn't have to override `model()` method.

*Changes*
- `CActiveRecord::model()` signature changed; `__CLASS__` replaced with `NULL`
- If `$className` is `NULL`, it will fall back to the name of the calling class


samdark:
Can't do it because 1.1 is still backwards compatible to 5.2.

adeelnawaz:
@samdark how about this one? #3897

PR #3897, opened 2015-06-22
CActiveRecord::model() defaults to called class

adeelnawaz:
*Purpose*
The `CActiveRecord::model()` method uses `__CLASS__` magical constant as a default fallback for class name. It must not use it as this constant is early static binding, hence it will always return *CActiveRecord*. Instead, the class must use PHP `get_called_class()` (PHP 5 >= 5.3) to get late static binding class name. So that every class that implements `CActiveRecord` doesn't have to override `model()` method.

*Changes*
- `CActiveRecord::model()` signature changed; `__CLASS__` replaced with `NULL`
- If `$className` is `NULL`, it will fall back to the name of the calling class
- Backward compatible with PHP 5<5.3

### DIFF
--- a/framework/db/ar/CActiveRecord.php
+++ b/framework/db/ar/CActiveRecord.php
@@ -388,6 +388,9 @@ abstract class CActiveRecord extends CModel
 	 */
 	public static function model($className=__CLASS__)
 	{
+                if($className===__CLASS__ && function_exists('get_called_class')){
+                    $className = get_called_class();
+                }
 		if(isset(self::$_models[$className]))
 			return self::$_models[$className];
 		else


klimov-paul:
This is unacceptable, because the bhavior of the `model()` method varies depending on PHP version.

adeelnawaz:
It is also unacceptable to override a method in every child class with exact same implementation. Before PHP 5.3 it was inevitable, but devs should not be forced to do so in 5.3 and beyond. Your concern is valid but that's what a graceful fallback does: when backward compatibility is not an option, it adapts a new behavior.

cebe:
This implementation can have unexpected side effects, better not implement it in the framework directly. If you want to avoid implementing this method in all model classes you can have your own base class implementing it. You should be fine if your app is covered by unit tests. I am quite sure this will break in some case.

adeelnawaz:
I'm already doing exactly that. Anyways, a pity I couldn't be of any help to others.

PR #3936, opened 2015-09-10
extractType modification

iConn:
Several slow queries dump problems working with unsigned int values. They become strings.
PHP_INT_MAX 9223372036854775807 (2^63-1) amd64
MySQL UNSIGNED INT 4294967295
MySQL BIGINT +/- 9223372036854775807
MySQL UNSIGNED BIGINT 18446744073709551615

### DIFF
--- a/framework/db/schema/mysql/CMysqlColumnSchema.php
+++ b/framework/db/schema/mysql/CMysqlColumnSchema.php
@@ -29,7 +29,7 @@ class CMysqlColumnSchema extends CDbColumnSchema
 			$this->type='double';
 		elseif(strpos($dbType,'bool')!==false)
 			$this->type='boolean';
-		elseif(strpos($dbType,'int')===0 && strpos($dbType,'unsigned')===false || preg_match('/(bit|tinyint|smallint|mediumint)/',$dbType))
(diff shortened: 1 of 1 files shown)


samdark:
I was already discussed several times. PHP integer size is less than MySQL's so it's not safe to represet types like you've proposed. Especially on 32 bit systems.

iConn:
what about:

elseif((strpos($dbType,'int')===0 && ( strlen(decbin(~0)) > 32  || strpos($dbType,'unsigned')===false)) || preg_match('/(bit|tinyint|smallint|mediumint)/',$dbType))

??

samdark:
That's quite a big inconsistency. Would be quite bad to migrate from 32 to 64 bits and get type break in, for example, external API which is consumed by type sensitive iOS or Android clients.

iConn:
... ¿iOS and Android serving PHP?, I do not get the point...
About migration from 32 to 64:
32b: strlen(decbin(~0)) === 32
64b: machine strlen(decbin(~0)) === 64
but... ok ;).  
brainstorm mode: off

samdark:
I mean REST API built with Yii and consumed with iOS/Android.

PR #3948, opened 2015-09-24
CHttpRequest: fix path info extraction

dmitrivereshchagin:
Add checks on whether value returned by `substr` is `false` or empty string. This is need for compatibility with PHP 7.


samdark:
Could you add a line for changelog?

samdark:
Thanks!

PR #3949, opened 2015-09-24
CConsoleCommand: control `copyFiles' interactivity

dmitrivereshchagin:
Add `overwriteAll` argument to `copyFiles` method which can be set to true in noninteractive commands.


samdark:
We can't change public signatures for backwards compatibility reasons.

samdark:
Probably in this case it's OK. Please add a line for changelog.

samdark:
Thanks!

PR #3960, opened 2015-10-06
Fix ignoring timezone in CFormatter::date with DateTimes

auburus:
Yii 1.1.16 introduced the ability to pass DateTime objects to CFormatter.

The problem is that `DateTime::getTimestamp()` ignores completely timezones, as it can be seen in the following code:

```php
$a = new DateTime('now', new DateTimeZone('UTC'));
$b = new DateTime('now', new DateTimeZone('America/Los_Angeles'));

if ($a->getTimestamp() === $b->getTimestamp()) {
    echo 'This line is printed' . PHP_EOL;
} else {
    echo 'This line isn't printed' . PHP_EOL;
}
```

using `DateTime::getTimestamp() + DateTime::getOffset()` solves the problem.

PD: I haven't found any test that currently evaluates this function, so adding a test case wasn't as easy as expected. Because of this, i provide this fix without any test. And being such an small fix, I haven't opened an issue.

### DIFF
--- a/framework/utils/CFormatter.php
+++ b/framework/utils/CFormatter.php
@@ -235,7 +235,7 @@ class CFormatter extends CApplicationComponent
 				return strtotime($time);
 		}
 		elseif (class_exists('DateTime', false) && $time instanceof DateTime)
-			return $time->getTimestamp();
+			return $time->getTimestamp() + $time->getOffset();
 		else
 			return (int)$time;
 	}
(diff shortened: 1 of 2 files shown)


samdark:
@cebe can you verify/merge this one?

framework/utils/CFormatter.php
```diff
@@ -235,7 +235,7 @@ protected function normalizeDateValue($time)
 				return strtotime($time);
 		}
 		elseif (class_exists('DateTime', false) && $time instanceof DateTime)
-			return $time->getTimestamp();
+			return $time->getTimestamp() + $time->getOffset();
```
cebe: A Unix timestamp is always UTC, there is no point in adding an offset here.

I see however that there may be a problem with this code in general so you might open an issue describing the problem you see when using the formatter.

PR #3976, opened 2015-10-16
Overriding applyJoin() to comply with the DELETE command syntax.

odalecne:
In SQL Server, when multiple tables are joined in a DELETE statement, the FROM clause in which JOIN clauses are declared must be preceded by an additional FROM clause (whose keyword is optional and usually omitted for readability and consistency) that specifies the target table name.
See syntax and examples at https://msdn.microsoft.com/en-us/library/ms189835.aspx.

### DIFF
--- a/framework/db/schema/mssql/CMssqlCommandBuilder.php
+++ b/framework/db/schema/mssql/CMssqlCommandBuilder.php
@@ -142,7 +142,7 @@ class CMssqlCommandBuilder extends CDbCommandBuilder
 	 */
 	public function applyJoin($sql,$join)
 	{
-		if(trim($join)!='')
+		if(trim($join)!=='')
 			$sql=preg_replace('/^\s*DELETE\s+FROM\s+((\[.+\])|([^\s]+))\s*/i',"DELETE \\1 FROM \\1",$sql);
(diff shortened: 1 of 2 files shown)


framework/db/schema/mssql/CMssqlCommandBuilder.php
```diff
@@ -134,6 +134,20 @@ public function createUpdateCounterCommand($table,$counters,$criteria)
...
 	/**
+	 * Alters the SQL to apply JOIN clause.
+	 * Overrides parent implementation to comply with the DELETE command syntax required when multiple tables are referenced.
+	 * @param string $sql the SQL statement to be altered
+	 * @param string $join the JOIN clause (starting with join type, such as INNER JOIN)
+	 * @return string the altered SQL statement
+	 */
+	public function applyJoin($sql,$join)
+	{
+		if(trim($join)!='')
```
samdark: `!==` should be used.

samdark:
Looks OK. Could you add changelog?

c-schmitz:
Patch works great here.., thank you!

samdark:
Merged. Thanks!

PR #3979, opened 2015-10-26
Update CSecurityManager.php

vortextangent:
Previously $validateEncryptionKey setting was not observed when set to false


samdark:
It's OK as is. We've decided key validation is mandatory a while ago. Marked property as deprecated.

PR #3980, opened 2015-10-28
Fix Redis cache with mb_strlen

Lexx918:
If you use a setting ..
```ini
mbstring.internal_encoding = UTF-8
mbstring.func_overload = 7
```
.. you can see the error ..
```
Redis error: ERR Protocol error: unbalanced quotes in request
```
in CRedisCache.php : 151
```php
case '-': // Error reply
    throw new CException('Redis error: '.$line);
```
The patch corrects the calculation of the lengths of strings.

### DIFF
--- a/framework/caching/CRedisCache.php
+++ b/framework/caching/CRedisCache.php
@@ -125,7 +125,7 @@ class CRedisCache extends CCache
 		array_unshift($params,$name);
 		$command='*'.count($params)."\r\n";
 		foreach($params as $arg)
-			$command.='$'.strlen($arg)."\r\n".$arg."\r\n";
+			$command.='$'.$this->lenBytes($arg)."\r\n".$arg."\r\n";
 
 		fwrite($this->_socket,$command);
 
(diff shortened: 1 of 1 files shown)


framework/caching/CRedisCache.php
```diff
@@ -177,6 +177,17 @@ private function parseResponse()
 	}
 
 	/**
+	 * Counting amount of bytes in a string.
+	 *
+	 * @param string $str
+	 * @return int
+	 */
+	private function lenBytes($str)
```
samdark: There's already such method in a helper. Could be used as `StringHelper::byteLength($string)`.

Lexx918: In yii2. But I use 1.1.16.

samdark: Oh. I'm not focused much today. Sorry.

Lexx918: Бывает (:

samdark:
Merged with some adjustments. Thanks!

PR #3982, opened 2015-10-29
Rearranged current() calls to comply with PHP strict standards.

odalecne:
Only variables should be passed by reference.


samdark:
Thanks!

Issue #3616, opened 2014-11-05
SQL Server throws exception when regenerating non-existent session


brentjanderson:
SQL Server 2012 SP 2 (and likely other versions of SQL server) rejects YiiSession table insertions in CDbHttpSession.php on line 129 (https://github.com/yiisoft/yii/blob/master/framework/web/CDbHttpSession.php#L129)

The null byte it attempts to insert (0x) with the single quotes results in an exception:

```
Implicit conversion from data type varchar to varbinary(max) is not allowed. Use the CONVERT function to run this query.
```

One potential solution would be to provide a default value for the data column in the YiiSession schema, and then use the following for SQL Server's schema:

```
[data] [varbinary](max) NOT NULL DEFAULT CONVERT(VARBINARY(MAX), '')
```

samdark:
Won't be fixed in 1.1.

PR #3984, opened 2015-11-02
CDbHttpSession Null Data for Regeneration

mkaufmaner:
Bug #3616: Fixed `CDbHttpSession` SQL Server exception when regenerating non-existent session by storing NULL instead of an empty string.


samdark:
Won't be fixed in 1.1.

PR #3991, opened 2015-11-19
Fixes the requires paths

cauealves:
Hello,

When I use HHVM and pre-compiling will get the following warns:

`Unable to stat file /yii/framework/gii/components/Pear/Text/Diff/Renderer/Text/Diff/Renderer.php`

`Unable to stat file /yii/framework/gii/components/Pear/Text/Diff/Text/Diff.php`

`Unable to stat file /yii/framework/gii/components/Text/Diff.php`

`Unable to stat file /yii/framework/gii/components/Text/Diff/Renderer.php`

`Unable to stat file /yii/framework/gii/components/Text/Diff/Renderer/inline.php`

And I resolved with this commit.

### DIFF
--- a/framework/gii/components/Pear/Text/Diff/Renderer/inline.php
+++ b/framework/gii/components/Pear/Text/Diff/Renderer/inline.php
@@ -14,7 +14,7 @@
  */
 
 /** Text_Diff_Renderer */
-require_once 'Text/Diff/Renderer.php';
+require_once '../Renderer.php';
 
 /**
  * "Inline" diff renderer.
--- a/framework/gii/components/Pear/Text/Diff/Renderer/unified.php
+++ b/framework/gii/components/Pear/Text/Diff/Renderer/unified.php
@@ -16,7 +16,7 @@
  */
 
 /** Text_Diff_Renderer */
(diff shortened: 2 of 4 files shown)


samdark:
Have you verified that Gii works with regular PHP after these changes?

cauealves:
Hello samdark, how are u?

With simple commands yes, but there are tests that I could be running?

cebe:
the paths are relative to defined php include paths. Your changes do not look correct to me, at least I do not see how these should work.

cauealves:
Hello cebe,

Thanks for comment, it's true.
Can you see a way how we can to solve this issue?

cebe:
can you specify different include_path when using precompiling?

cauealves:
In HHVM documentation I have not found anything about it...

cebe:
http://docs.hhvm.com/manual/en/ini.core.php#ini.include-path

set this to contain all directories that are included by Yii using statements like `Yii::import('gii.components.Pear.*');` Everything that ends with `*` will result in the path to be added to the include_path. So if you need these path before running yii code you have to manually set them.

PR #3996, opened 2015-12-02
Remove base path from getId

tdbui22:
My production setup uses a new directory per deployment and then creates a symlink back to the webroot.

The problem I'm having is this method is used to create the session key. When I deploy, the realpath actually changes which causes the app to generate a new session key which results in my users getting logged due to lost of previous session.


samdark:
Set your app ID in config.

PR #4004, opened 2016-01-15
Fix #3476: Database sessions with Postgres did not work properly

### LINKED ISSUE #3476: use session with postgresql
yii version 1.1.15

my config :
        'db'=>array(
            'class'=>'CDbConnection',
            'connectionString'=>'pgsql:host=localhost;port=5432;dbname=comm',
            'username'=>'postgres',
            'password'=>'123456',
        ),
		'session' => array (
            'class' => 'system.web.CDbHttpSession',
            'connectionID' => 'db',
            'sessionTableName' => 'session_table',
        ),

table structure for session_table:
		CREATE TABLE "public"."session_table" (
		"id" char(32) COLLATE "default" NOT NULL,
		"expire" int4,
		"data" bytea
		)
		WITH (OIDS=FALSE);
		
yii can't read session data, i changed the bytea into varchar, problem solving!

GuillaumeSmaha:
In the case when the session contains a class, the fix from https://github.com/yiisoft/yii/commit/74934b0a5856af5052b8127920622c9826862a26 doesn't work because $db->quoteValue truncate the value. 
PDO::quote for postgresql calls the function PQescapeStringConn which stops to parse the string at the first \0 (EOL) and PHP uses the char \0 as a mark to serialize class. (http://www.postgresql.org/docs/9.4/interactive/libpq-exec.html)

My fix calls PDO::quote with the parameter PDO::PARAM_LOB which calls PQescapeByteaConn instead of PQescapeStringConn and so convert_to is useless.
#3476

### DIFF
--- a/framework/db/CDbConnection.php
+++ b/framework/db/CDbConnection.php
@@ -583,6 +583,25 @@ class CDbConnection extends CApplicationComponent
 	}
 
 	/**
+	 * Quotes a string value for use in a query.
+	 * @param string $str string to be quoted
+	 * @param integer $quoteParam Parameter for PDO::quote function.
+	 * @return string the properly quoted string
+	 * @see http://www.php.net/manual/en/function.PDO-quote.php
+	 */
+	public function quoteValueExtended($str, $quoteParam)
+	{
+		if(is_int($str) || is_float($str))
+			return $str;
+
+		$this->setActive(true);
+		if(($value=$this->_pdo->quote($str, $quoteParam))!==false)
+			return $value;
+		else  // the driver doesn't support quote (e.g. oci)
+			return "'" . addcslashes(str_replace("'", "''", $str), "\000\n\r\\\032") . "'";
+	}
+
+	/**
 	 * Quotes a table name for use in a query.
 	 * If the table name contains schema prefix, the prefix will also be properly quoted.
 	 * @param string $name table name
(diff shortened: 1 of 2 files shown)


framework/db/CDbConnection.php
```diff
@@ -567,16 +567,17 @@ public function getLastInsertID($sequenceName='')
 	/**
 	 * Quotes a string value for use in a query.
 	 * @param string $str string to be quoted
+	 * @param integer $quoteParam Parameter for PDO::quote function. Default : PDO::PARAM_STR
 	 * @return string the properly quoted string
 	 * @see http://www.php.net/manual/en/function.PDO-quote.php
 	 */
-	public function quoteValue($str)
+	public function quoteValue($str, $quoteParam = PDO::PARAM_STR)
```
samdark: Public method signature change doesn't look right.

cebe:
we may introduce the array syntax that we have in yii2.

framework/db/CDbConnection.php
GuillaumeSmaha: I used the same method signature as PDO::quote(string, integer). I don't understand what does not look right ?

samdark: I mean you're changing existing public method signature which means backwards compatibility break.

GuillaumeSmaha:
@samdark Ok, I create a new function to avoid backwards compatibility break.

samdark:
@cebe do you plan to work on it?

cebe:
Merged with small adjustments, thank you!

PR #4013, opened 2016-01-28
Use both baseUrl and basePath in loading scripts

matjack1:
I'm not sure if you are interested in this, but I've found it useful when I had to add a client script in my own CDN pointing inside the assets folder.

I know that Yii1 is closed for enhancements, but I thought this was a very small change that might benefit someone else :)

### DIFF
--- a/framework/web/CClientScript.php
+++ b/framework/web/CClientScript.php
@@ -541,6 +541,8 @@ class CClientScript extends CApplicationComponent
 			if($baseUrl==='' || $baseUrl[0]!=='/' && strpos($baseUrl,'://')===false)
 				$baseUrl=Yii::app()->getRequest()->getBaseUrl().'/'.$baseUrl;
 			$baseUrl=rtrim($baseUrl,'/');
+			if(isset($package['basePath']))
+				$baseUrl=$baseUrl.Yii::app()->getAssetManager()->publish(Yii::getPathOfAlias($package['basePath']));
 		}
 		elseif(isset($package['basePath']))
 			$baseUrl=Yii::app()->getAssetManager()->publish(Yii::getPathOfAlias($package['basePath']));


cebe:
could you explain a bit more what it does and how it works?

matjack1:
Yes sure, sorry, the idea is to use it like so:
```
'clientScript' => [
	'packages' => [
		'application' => [
			'basePath' => 'application.assets',
			'baseUrl' => getenv( 'CDN_URL' ),
			'js' => ['js/application.min.js'],
		],
	],
],
```
In order to be able to specify both `basePath` and `baseUrl`, otherwise currently if I specify `baseUrl`, `basePath` is ignored. The PR allow us to reference an asset managed by Yii through a CDN and use the `clientScript` dependencies and inclusion methods (for example).

lexeo:
Hello. Did you try to do it this way?
```
// protected/config/main.php

// ...
'assetManager' => array(
    'basePath' => 'application.assets',
    'baseUrl' => 'https://cdn.example.com',
),
'clientScript' => array(
    'packages' => array(
        'zeroclipboard' => array(
            'depends' => array('jquery'),
            'basePath' => 'vendors.components.zeroclipboard',
            'js' => array('dist/ZeroClipboard.js'),
        ),
    ),
),
// ...
```

matjack1:
Hello @lexeo thanks for suggesting that, I've tried but it doesn't work here, I get the error:
`CAssetManager.basePath "application.assets" is invalid. Please make sure the directory exists and is writable by the Web server process.`

Is it working for you?

lexeo:
Sorry, my friend. I made a little mistake. 
At first, [alias](http://www.yiiframework.com/doc/guide/1.1/en/basics.namespace#root-alias) "application" refers to the [application's base directory](http://www.yiiframework.com/doc/guide/1.1/en/basics.application#application-base-directory);
Secondly, I used the alias instead of absolute path.
You need to define a valid path to assets directory. Please, [look here](http://www.yiiframework.com/doc/api/1.1/CAssetManager#setBasePath-detail) to understand how it works.
For example, use:
```php
 'basePath' => \Yii::getPathOfAlias('webroot.assets'),
```

matjack1:
Thanks a lot @lexeo! You cleared my mind about the issue, and your explaination made me delete a bunch of code that I wrote to handle our CDN in the views. Awesome, thanks again, I'll close the PR then, it's not needed.

PR #4015, opened 2016-02-13
Inherit "disabled" attribute in internally rendered hidden fields in checkbox, radio, etc.

rob006:
This PR fix bug, when CHtml render active hidden filed for disabled checbox or radio buttons.

```php
echo CHtml::activeCheckBox($model, 'active', array('disabled' => true));
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,7 @@
 Version 1.1.18 under development
 --------------------------------
 
-- no changes have been made yet.
+- Bug #4015: Fixed bug with missing "disabled" attribute in internally rendered hidden fields (rob006)
 
 Version 1.1.17 January 13, 2016
 ------------------------------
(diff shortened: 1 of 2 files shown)


cebe:
What exactly is the "bug" or problem in this case?

cebe:
also need a CHANGELOG entry.

rob006:
Disabled input does not send POST data, they are ignored by browser on form submit. If we want render disabled checkbox, we expect so specified attribute will not be available in `$_POST['MyModel']`. But right now `CHtml` render:
```html
<input id="ytMyModel_active" type="hidden" value="0" name="MyModel[active]" />
<input disabled="disabled" name="MyModel[active]" id="MyModel_active" value="1" checked="checked" type="checkbox" />
```
Hidden input is not disabled, so it will be send in POST. If we don't block modify this attribute in validation rules, it also set wrong value (it always send '0' event if actual value i '1').

rob006:
I added changelog entry.

samdark:
Looks fine.

cebe:
Merged, thank you!

rob006:
I have not tested this, but this bug may also affect Yii 2.

cebe:
yes, same issue exists in 2.0, created an issue: https://github.com/yiisoft/yii2/issues/10843

PR #4021, opened 2016-03-04
use alias from criteria in CDbCommandBuilder if it's present

askobara:


samdark:
Won't be fixed in 1.1. Thanks for contribution though.

PR #4046, opened 2016-05-25
Allow absolute URLs to start with '//' instead of just 'http' - e.g. …

ksowa:
…'//google.com' is valid absolute URL

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 Version 1.1.18 under development
 --------------------------------
 
+- Enh #4049: Added '//' as a proper beginning of absolute URL in createAbsoluteUrl() method (ksowa)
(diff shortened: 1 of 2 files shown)


ksowa:
I ran into that issue with my project. For console application I defined 
`$_SERVER['SERVER_NAME'] = '//domain.com';`
for absolute URLs creation - e.g. for activation emails or internal messages system, as we allow both HTTP and HTTPS URLs. The '//' works with all protocols. 

With current Yii I ended up with http:////domain.com//domain.com/path 

With the commit I made it will properly create //domain.com/path URL.

samdark:
Makes sense. Could you add a line for changelog?

ksowa:
Added.

samdark:
Thanks!

Issue #4043, opened 2016-05-23
CJavaScript::quote doesn't quote \u2028, which breaks JS


marrch-caat:
If some text contains unicode character 'LINE SEPARATOR' (U+2028), it isn't quoted, and then JS code containing this character inside a string constant is broken (Uncaught SyntaxError: Unexpected token ILLEGAL). Reproduced in (at least) all the last Chrome and Firefox builds.

samdark:
How to reproduce the issue exactly?

marrch-caat:
Well, I cannot provide a full code sample, but if you take any text containing the \u2028 symbol, and insert its content to JS processing it via CJavaScript::quote, like this:
...
$text = CJavaScript::quote($text);
?>
<script type='text/javascript'>
var text = '<?=$text?>';
...
- you'll get the JS broken on execution, producing JS error Uncaught SyntaxError: Unexpected token ILLEGAL.

SamMousa:
@samdark What about:
````
public static function quote($js,$forUrl=false)
{
    $encoded = substr(json_encode($js), 1, -1);
    return $forUrl ? strtr($encoded, array('%' => '%25')) : $encoded;
}
````

samdark:
@SamMousa that actually works :)

samdark:
But only partially. It breaks on double quotes and single quotes.

SamMousa:
Will look into it and make a PR.

samdark:
Things to consider:

- https://www.inanimatt.com/php-output-escaping.html
- http://www.texelate.co.uk/blog/post/131-convert-php-strings-for-use-as-javascript-strings/
- https://github.com/twigphp/Twig/blob/master/test/Twig/Tests/escapingTest.php — tests.
- https://github.com/zendframework/zend-escaper/blob/master/src/Escaper.php#L178 — Zend variant of solving it.

samdark:
Also there's nothing about it in Yii 2 so  when you do `registerJavaScript()` with strings from the issue here it behaves the same way...

SamMousa:
Hmm I received an email about it breaking with quotes, but can't reproduce that. Is that why you removed it?

SamMousa:
I've made a PR

SamMousa:
https://framework.zend.com/manual/2.3/en/modules/zend.escaper.escaping-javascript.html
I cannot reproduce the vulnerability detailed here, it seems to me that this depends on the version of PHP?

samdark:
[...]

PR #4047, opened 2016-05-27
fixes: #4043

tomtomsen:
refs #4043

Based on http://stackoverflow.com/questions/14179377/how-to-replace-escape-u2028-or-u2029-characters-in-php-to-stop-my-jsonp-api-br i have tried to fix this issue.

additional to U+2028 character (Unicode line separator) it replaces the U+2029 character (Unicode paragraph separator).

the only way, i was able to test it, was to add following lines to a Controller and check the output for javascript errors. (copied "rocks" from https://web.archive.org/web/20150502034803/http://timelessrepo.com/json-isnt-a-javascript-subset)

``` php
$text = 'ro cks!
    test';
$text = CJavaScript::encode($text);

Yii::app()->clientScript->registerScript('invalid-character-js', 'var x = ' . $text . '; console.log(x);');
```

I don't thinks its possible to write unit tests for it.

``` php
'ro cks!
    test';
```

will be replaced with 

``` php
'ro\u2028cks!\n\ttest'
```

### DIFF
--- a/framework/web/helpers/CJavaScript.php
+++ b/framework/web/helpers/CJavaScript.php
@@ -28,9 +28,14 @@ class CJavaScript
 	public static function quote($js,$forUrl=false)
 	{
 		if($forUrl)
-			return strtr($js,array('%'=>'%25',"\t"=>'\t',"\n"=>'\n',"\r"=>'\r','"'=>'\"','\''=>'\\\'','\\'=>'\\\\','</'=>'<\/'));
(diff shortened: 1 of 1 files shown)


samdark:
Could be something like https://github.com/zendframework/zend-escaper/blob/master/src/Escaper.php but I haven't verified that.

samdark:
Closing since solution doesn't produce expected results.

PR #4065, opened 2016-07-12
Fixes #4064: CHtml::beginForm() produces wrong HTML when using GET and an anchor

### LINKED ISSUE #4064: CHtml::beginForm() produces wrong HTML when using GET and an anchor
This bug will appear appear with the default Yii settings, whenever a GET form contains both parameters in the URL and an anchor. Enabling pretty URLs (URL rewriting) makes this bug less frequent.

The following code will produce a wrong HTML:

```
echo CHtml::beginForm(['site', '#' => 'anchor'], 'get');
echo CHtml::endForm();
```

The resulting HTML will have an `input` tag with the attributes `name="r" value="site#anchor"` instead of  `value="site"`. Submitting the form will hence fail with an "unknown action" error.

mytskine:
CHtml::beginForm() was splitting the action URL by &
in order to extract the GET parameters.
When an anchor was appended to the URL, it was mixed with the last GET parameter.

An existing unit test was extended to cover this case.


samdark:
Merged. Thanks!

PR #4069, opened 2016-07-22
Fix CRedisCache::executeCommand() error when the previous connection timeout exception was catched

taobig:
try {
            Yii::app()->redisCache->executeCommand('INCR', ["key"]);
     } catch (Exception $e) {
            //catch the exception
}
try {
        Yii::app()->redisCache->executeCommand('GET', ["key"]);//[error] [php] fwrite() expects parameter 1 to be resource, boolean given 
      } catch (Exception $e) {
}


samdark:
Could you add a line for changelog?

PR #4076, opened 2016-08-04
Small change to CCaptcha.php 

lacrioque:
to prevent accidental duplication of captcha refresh button.


samdark:
Why would it duplicate?

samdark:
Closed because there's no explanation for a long time.

Issue #4078, opened 2016-08-09
AR  __isset function ,why return false here？  


greedying:
```php
    public function __isset($name)
    {
        if(isset($this->_attributes[$name]))
            return true;
        elseif(isset($this->getMetaData()->columns[$name]))
            return false;
        elseif(isset($this->_related[$name]))
            return true;
        elseif(isset($this->getMetaData()->relations[$name]))
            return $this->getRelated($name)!==null;
        else
            return parent::__isset($name);
    }
```
  why

```php
        elseif(isset($this->getMetaData()->columns[$name]))
            return false;
```

when         

```php
  public function hasAttribute($name)
    {
        return isset($this->getMetaData()->columns[$name]);
    }
```

thanks

cebe:
because that check is made after testing whether an attribute value is present. The semantics of isset are to be true when a value exists and it is not null.

The first check tests that condition for an attribute: `        if(isset($this->_attributes[$name]))`.
If no attribute value exists but the property is an attribute, it is not set. so we return false here.

PR #4079, opened 2016-08-09
where AR has named attribute, function __isset return true

greedying:
where AR has named attribute, function __isset should return true in my opinon


cebe:
the code is correct, see my explanation at https://github.com/yiisoft/yii/issues/4078#issuecomment-238481567

PR #4096, opened 2016-10-10
 query to database is executed twice

gekannt:
We need do something with this.  If we have a time consuming query which takes e.g. 10 seconds, we are end up with running it twice, I.e. user will wait for answer 20 seconds instead of 10. Run into this , commenting this help to speed up things at least by two. 

I can't see any severe problems to what it can lead. Do you?


gekannt:
ok, it is not the reason

PR #4098, opened 2016-10-17
add labels languages by default

kdchaires:
add labels languages by default, I hope the change is accepted


samdark:
No, it won't be accepted. First of all, [Yii 1.1 is in security-maintenance only mode](http://www.yiiframework.com/news/90/update-on-yii-1-1-support-and-end-of-life/). Second, it changes existing behavior.

PR #4099, opened 2016-10-20
Fix CHtml::value returning $defaultValue

damien-biasotto:
… when `$attribute` == 'array[like]'

We've just discovered a small regression in `CHtml::value` method.

If `$attribute` is not in a dot syntax but in an array syntax (like `product[name]`) and `$model` is an object it won't display corresponding value (even if it exists) but `$defaultValue` instead.


damien-biasotto:
Wrong fix. I will reopen the PR with a proper fix.

PR #4106, opened 2016-11-22
Make composer package lighter

b1rdex:
No one needs demos and concatenated stuff through composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,9 +69,7 @@
 	],
 	"archive": {
 		"exclude": [
-			"demos",
-			"requirements",
-			"framework/yiilite.php"
+			"demos"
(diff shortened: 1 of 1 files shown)


samdark:
Agree about demos but don't agree about the rest.

b1rdex:
Ok, let's remove just demos.

samdark:
Thanks!

b1rdex:
@samdark what do you think about removing `build`, `docs` and `tests` folders aswell? Almost 10MB.

samdark:
They are already removed via `.gitattributes`: https://github.com/yiisoft/yii/blob/master/.gitattributes

PR #4108, opened 2016-11-25
add comment to CWebuser getStateKeyPrefix()

aXen:
Using "CWebuser getStateKeyPrefix" should be careful, when deployed in a cluster environment


samdark:
That's too specific to be there.

samdark:
Thanks for contributing though.

PR #4112, opened 2016-12-05
PHP 7.1 compatibility

cebe:
Mainly build environment adjustments to make sure tests run fine. Additionally some fixes in TextHighlighter:

- adjusted `TextHighlighter` to load fine with constants.
- adjusted `TextHighlighter` to work the same on PHP 7.1. as on other versions by introducing an explicit cast to array() which is not allowed to be implicit anymore in PHP 7.1.

Is there an upstream repo of the TextHighlighter where we could contribute these back?

### DIFF
--- a/framework/vendors/TextHighlighter/Text/Highlighter.php
+++ b/framework/vendors/TextHighlighter/Text/Highlighter.php
@@ -46,7 +46,9 @@ if (!defined('HL_NUMBERS_LI')) {
 /**
  * for our purpose, it is infinity
  */
-define ('HL_INFINITY',      1000000000);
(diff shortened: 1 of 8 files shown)


framework/vendors/TextHighlighter/Text/Highlighter.php
```diff
@@ -46,7 +46,9 @@
 /**
  * for our purpose, it is infinity
```
cebe: :laughing:

samdark:
TextHighlighter is abandoned: http://pear.php.net/package/Text_Highlighter

cebe:
Okay, so it should be fine to ship a modified version.

PR #4123, opened 2017-01-16
croatian translation

TheAdnan:
Added the Croatian translation.


TheAdnan:
The Bosnian translation was edited because it had many grammatical errors and was in the wrong dialect in some places.

samdark:
Thanks!

PR #4130, opened 2017-01-30
Update CComponent.php

devcommiter:
A little protection of the eval() usage in the evaluateExpression method.


samdark:
Why closing?

framework/base/CComponent.php
```diff
@@ -609,7 +609,14 @@ public function evaluateExpression($_expression_,$_data_=array())
 		if(is_string($_expression_))
 		{
 			extract($_data_);
-			return eval('return '.$_expression_.';');
+			try
+			{
+				return @eval('return ' . $_expression_ . ';');
```
cebe: why did you add the `@` here?

devcommiter: For suppressing errors from eval() function. I think in this case we will simply parse a necessary exception or get clear false without any additional messages. Similar solution you can see \CChoiceFormat::evaluate or in the \CAuthManager::executeBizRule or \CTimestampBehavior::getTimestampByAttribute

samdark: That changes behavior to be silent if PHP version is lower than 7. I don't think we should change it.

samdark:
Looks good to me. Please add a line for changelog.

devcommiter:
Hi, there is already exist changelog's comment.
`Bug #4130: Fixed PHP 7 added an interception of the ParseError exception for the eval() usage (devcommiter)`

samdark:
Merged. Thanks!

PR #4131, opened 2017-01-30
Update CPropertyValue.php

devcommiter:
Fixed an error in the eval() usage in the ensureArray method.

### DIFF
--- a/framework/utils/CPropertyValue.php
+++ b/framework/utils/CPropertyValue.php
@@ -113,8 +113,14 @@ class CPropertyValue
 			$len = strlen($value);
 			if ($len >= 2 && $value[0] == '(' && $value[$len-1] == ')')
 			{
-				eval('$array=array'.$value.';');
-				return $array;
+				try
+				{
+					return @eval('return array' . $value . ';');
+				}
+				catch (ParseError $e)
+				{
+					return array();
+				}
 			}
 			else
 				return $len>0?array($value):array();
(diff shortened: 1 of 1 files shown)


samdark:
Why closing?

devcommiter:
Oh, it's my fail, sorry. I'll re-create requests.

framework/utils/CPropertyValue.php
```diff
@@ -113,8 +113,14 @@ public static function ensureArray($value)
 			$len = strlen($value);
 			if ($len >= 2 && $value[0] == '(' && $value[$len-1] == ')')
 			{
-				eval('$array=array'.$value.';');
-				return $array;
+				try
+				{
+					return @eval('return array' . $value . ';');
```
cebe: why did you add the `@` here?

devcommiter: For suppressing errors from eval() function. I think in this case we will simply parse a necessary exception or get clear false without any additional messages. Similar solution you can see \CChoiceFormat::evaluate or in the \CAuthManager::executeBizRule or \CTimestampBehavior::getTimestampByAttribute

samdark: That changes behavior to be silent if PHP version is lower than 7. I don't think we should change it.

samdark:
Looks good to me now. Please add a like for changelog.

@cebe could you take a brief look again?

devcommiter:
Please, do not spend time for checking this request, I'd better add more changes around eval() into 'patch-2'(https://github.com/yiisoft/yii/pull/4130) and this change for CPropertyValue.php also.
(p.s. I don't want to add many lines into the changelog with the same meaning). What do you think?

samdark:
OK.

PR #4132, opened 2017-01-30
Improved the eval() usage compatibility with a PHP7.0

devcommiter:


samdark:
Html purifier is a third party library so preferably it should be enhanced there and then updated in the 1.1 repo.

PR #4133, opened 2017-02-01
Fixes for the func_get_args() usage for PHP7

devcommiter:


cebe:
can you please describe the problem you are fixing here?

devcommiter:
Hi, of course.

As you can see on this URL [http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.other.func-parameter-modified](url) the behavior of the func_get_arg() and func_get_args() has been changed in the PHP7.0 and as a result you won't get the arguments which you are expecting, IF you had modified the input parameters of the function in the source code before you call func_get_arg() or func_get_args().
In the PHP5 you were getting untouched version of the input parameters by calling func_get_arg() or func_get_args(), but not in the PHP7.

In the CClientScript.php this problem fixed easy - I simply moved the func_get_args() to the upper position than the strings where the input parameters were modified. I skept the functions where this problem is not appearing.

In the YiiBase.php, I decided to update a bit a source code of the function.

All these modification are backward compatibility both with PHP5 and PHP7.

framework/YiiBase.php
```diff
@@ -179,6 +179,9 @@ public static function getFrameworkPath()
 	 */
 	public static function createComponent($config)
 	{
+		$intArgsNum = func_num_args();
```
samdark: Nothing changed about `func_num_args()` in PHP 7. The change is not necessary.

samdark:
Looks good to me now. Please add a like for changelog file and I'll merge it.

devcommiter:
Changelog: Fixed a wrong usage of the func_get_arg() and func_get_args() functions. See URL [http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.other.func-parameter-modified](url) for more details.

samdark:
I've meant https://github.com/yiisoft/yii/blob/master/CHANGELOG

samdark:
Merged. Thank you!

PR #4146, opened 2017-03-19
Yii Namespaced Demo

Fractal-Idea:


cebe:
Why are you adding these files? Any comment about the purpose of this PR? Looks like these should rather be in a separate repository to me.

Fractal-Idea:
As [demos/namespace/README.md](https://github.com/Fractal-Idea/yii/blob/master/demos/namespace/README.md)

this demo is totally based on Yii 1.x framework, so it's not a separate repository.
this demo show how to use `namespace` based on Yii 1.x framework
multiple sites shared `lib` and `Yii framework`
multiple entries in one site  (such index / admin / manager / etc. )
namespaced model generator integrated to `gii`
support environment-based config, common config data write once

[demos/namespace/lib/models/BaseActiveRecord.php](https://github.com/Fractal-Idea/yii/blob/master/demos/namespace/lib/models/BaseActiveRecord.php)
**BaseActiveRecord** : if table schema changed, regenerate model code will erase custom code. extend model class for custom code, if table schema changed we lost nothing when regenerate model code. 

**BaseActiveRecord::Model** : You should use class name for ExtendModelClass::model('ExtendModelClass'), otherwise CActiveRecord::model will wrong. BaseActiveRecord::model make it simple & clear : `ExtendModelClass::model()`

**BaseActiveRecord::getAlias** : simple way to fetch aliased sql expression result
```SQL
-- demos/namespace/site_demo/entry/dev/controllers/SiteController.php line 12
SELECT *, FORMAT((num_a+num_b+num_c)/3, 2) AS $$avg FROM tb_demo
```
```PHP
// demos/namespace/site_demo/entry/dev/views/site/index.php line 53
<td><?php echo($demo->id)?></td>
<td><?php echo($demo->num_a)?></td>
<td><?php echo($demo->num_b)?></td>
<td><?php echo($demo->num_c)?></td>
<td><?php echo($demo->getAlias('$$avg'))?></td>
```
**BaseActiveRecord::instantiate** : only alias name start with `$$` saved for fetch, others are skipped.

cebe:
This is a nice implementation, but it is yours and not something we (the Yii team) would want to maintain, also considering the fact that Yii 1.1 is in maintenance mode now.
Please create your own repository for it. You may create a wiki page and/or an extension page on the website to share it with others. Thanks!

Fractal-Idea:
There are better ways to implement these features by directly modify yii framework main code:
- Modify CActiveRecord::model
- Add CApplication::config
- Add yii/gii/genertors/namespaceModel

I know the fact that Yii 1.1 is in maintenance mode now. I do not modify any file in main code of yii framework. 
Only create demo for namespace use guide, and limit all changes (extend AR, namespace model generator for gii, env-based config, etc.) in the demo project.

PR #4158, opened 2017-05-26
Reworked model error escaping to escape on output rather than in validators directly

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


cebe:
what if application code relies on errors being encoded?

samdark:
It should not since context may vary. I'll add a note for UPGRADE.

samdark:
https://github.com/yiisoft/yii/commit/68fe34dbd0d2c886443c5af25f345c2498b0288f

PR #4165, opened 2017-07-15
Update: Show erros if can't start session on php7

georgezim85:
Taking off this @ will help people to know what is happening, when they do not have access permission read/write on xampp's temp directory or any other problem that do not allow them to start session.


samdark:
Tests are failing with the code changing...

cebe:
also this is neither a PHP7 compatibility nor a security issue. Not going to make that change in Yii 1.1. http://www.yiiframework.com/news/90/update-on-yii-1-1-support-and-end-of-life/

PR #4168, opened 2017-08-02
Session data not updating after login

csears123:
Updating from PHP 5.6 to PHP 7.1 introduced an issue where the session data no longer was updating, preventing the user from logging in. After debugging it was discovered the queryScalar in the readSession function was returning NULL from the empty record/row in the database with the passed in session ID.  The return statement has a strict comparison to check for boolean false, and did not account for a NULL value being returned from the empty "data" column in the database.  Changing the return statement to also check for a NULL resolves this issue, and will allow the session data to be updated correctly.  The config for the session is using CDbHttpSession class with a custom session name, and a CDbConnection that connects to IBM DB2 database.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no


samdark:
Merged. Thanks!

Issue #4172, opened 2017-08-25
No way to override PostgreSQL DEFAULT_SCHEMA


faisalkaleem:
Hi,
I have switched an application from MySQL to PostgreSQL. I observed that by default it uses "public" schema in **db.schema.pgsql.CPgsqlSchema** and there is now way to override this.
I observed that if we take private variable rather than a constant **DEFAULT_SCHEMA** then we'll have the ability override that.

samdark:
Since it's not PHP 7 compatibility issue or security issue, it won't be addressed. See http://www.yiiframework.com/news/90/update-on-yii-1-1-support-and-end-of-life/

PR #4173, opened 2017-08-25
Ability to override the default schema for PostgreSQL

faisalkaleem:
https://github.com/yiisoft/yii/issues/4172


samdark:
Since it's not PHP 7 compatibility issue or security issue, it won't be addressed. See http://www.yiiframework.com/news/90/update-on-yii-1-1-support-and-end-of-life/

PR #4183, opened 2017-12-26
docs [docs/guide/topics.i18n.txt] fix URL to "Language Plural Rules"

egorrishe:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

I've not changed code, only docs.
Replaced old deprecated link:
http://unicode.org/repos/cldr-tmp/trunk/diff/supplemental/language_plural_rules.html 
with new one:
http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html


samdark:
Merged. Thanks!

Issue #4184, opened 2018-01-04
PHP 7.2 - "ini_set(): Headers already sent." error in CHttpSession


softark:
### What steps will reproduce the problem?

Perform CHttpSessionTest on PHP 7.2, then you'll get the following error:

```
CHttpSessionTest::testSetGet
ini_set(): Headers already sent. You cannot change the session module's ini settings at this time.

/home/travis/build/softark/yii/framework/web/CHttpSession.php:273
/home/travis/build/softark/yii/framework/base/CComponent.php:152
/home/travis/build/softark/yii/framework/YiiBase.php:220
/home/travis/build/softark/yii/framework/base/CModule.php:393
/home/travis/build/softark/yii/framework/base/CModule.php:103
/home/travis/build/softark/yii/tests/framework/web/CHttpSessionTest.php:6
/home/travis/build/softark/yii/tests/framework/web/CHttpSessionTest.php:28
/home/travis/build/softark/yii/vendor/phpunit/phpunit/phpunit:52

```
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.19
| PHP version      | 7.2
| Operating system | ? (travis-ci)

daniel1302:
Fixed here https://github.com/yiisoft/yii/pull/4191/commits/0be5fbb5b359065beb2770feb262a4a8f689a309

PR #4191, opened 2018-01-28
Filters  for YiiBase::autoloader autoloader

daniel1302:
Add filters  for YiiBase::autoloader autoloader that allows to skip autoloading for specific classes.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

I will try to describe problem that i found in my project and I cannot fix it. 
I have legacy project in PHP5.6. This is enormous application. We cannot migrate to Yii2 at the moment.

I am trying to upgrade PHP to 7.2. We use Smarty 3.1.12 and it use a lot of deprecated and removed in php 7.2 stuff. I upgraded Smarty to 3.1.31(latest). 

We have got set YiiBase::$enableIncludePath to true in our project.
Smarty is extendable. Plugin for smarty is function. I use(https://github.com/yiiext/smarty-renderer). The latest version of smarty looking for a builtin class if it will find new tag in template file. If it not found it looking for a plugin function file. But if Smarty autoloader does not find plugin class Yii autoloader trying to find a class file. 

I will use example with plugin for meta tag {meta ...}
When YiiBase::$enableIncludePath === true autoloader throws error 
```
include(Smarty_Internal_Compile_Meta.php): failed to open stream: No such file or directory 
```

Second one solution is map all class for smarty via YiiBase::$classMap to existing file eg. yii.php. But it is awful solution (performance and code clean).

There are more info about problem. 
https://pastebin.com/RybZ0mdT

I belive a lot of people use Yii1.1 in their projects and it is hard do upgrade framework version to 2. We have upgrade php due to security issues, so my solution should be good way to avoid problem with autoloaders. In my code are not collistions with backward compability.

### DIFF
--- a/framework/YiiBase.php
+++ b/framework/YiiBase.php
@@ -53,12 +53,13 @@ defined('YII_ZII_PATH') or define('YII_ZII_PATH',YII_PATH.DIRECTORY_SEPARATOR.'z
  */
 class YiiBase
 {
-    /**
-     * @var array filters for autoloading mechanism.
-     * It should be callable. For callable function autoloader pass className.
-     * If filter function returns true Yii autoloader will be skipped.
-     */
-    public static $autoloaderFilters=array();
+	/**
+	 * @var array filters for autoloading mechanism.
+	 * It should be callable. For callable function autoloader pass className.
+	 * If filter function returns true Yii autoloader will be skipped.
+	 * @since 1.1.20
+	 */
+	public static $autoloaderFilters=array();
 	/**
 	 * @var array class map used by the Yii autoloading mechanism.
 	 * The array keys are the class names and the array values are the corresponding class file paths.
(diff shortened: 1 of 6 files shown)


daniel1302:
Fixed this issue too: https://github.com/yiisoft/yii/issues/4184

softark:
Wow, all checks have passed!!
I think it's very nice to have Yii 1.1.x that runs on PHP 7.2

BTW, you should respect the core framework code style
https://github.com/yiisoft/yii/wiki/Core-framework-code-style
> Code MUST use tabs for indenting, not spaces.

framework/YiiBase.php
```diff
@@ -53,6 +53,12 @@
  */
 class YiiBase
 {
+    /**
+     * @var array filters for autoloading mechanism.
+     * It should be callable. For callable function autoloader pass className.
+     * If filter function returns true Yii autoloader will be skipped.
```
samdark: `@since` needed.

daniel1302:
I fixed issues, Please force build, because for PHP 5.4 is an problem with build environment

samdark:
Done.

daniel1302:
@samdark I added entries in changelog :)
Thanks for aprove it :)

samdark:
Merged. Thank you!

PR #4203, opened 2018-04-26
add \CHttpSession::setCacheLimiter

b1rdex:
(cherry picked from commit b6306908cfcd59d84269739ea3bd845de78c8be1)

### DIFF
--- a/framework/web/filters/CHttpCacheFilter.php
+++ b/framework/web/filters/CHttpCacheFilter.php
@@ -193,7 +193,7 @@ class CHttpCacheFilter extends CFilter
 	{
 		if(Yii::app()->session->isStarted)
 		{
-		    Yii::app()->session->setCacheLimiter('public');
+			Yii::app()->session->setCacheLimiter('public');
 			header('Pragma:',true);
 		}
 		header('Cache-Control: '.$this->cacheControl,true);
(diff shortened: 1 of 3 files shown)


b1rdex:
Done.

samdark:
Merged. Thank you!

framework/web/filters/CHttpCacheFilter.php
```diff
@@ -193,7 +193,7 @@ protected function sendCacheControlHeader()
 	{
 		if(Yii::app()->session->isStarted)
 		{
-			session_cache_limiter('public');
+			Yii::app()->session->setCacheLimiter('public');
```
rob006: What is purpose of `session_cache_limiter()` here? This change broken `CHttpCacheFilter` in my case - for some reason I get `last-modified` header always set to mtime of entry script (which is obviously incorrect), and I'm unable to get 304 response from this filter.

samdark: These should be equal.

rob006: Disabling and reenabling session is definitely not transparent - I was also wondering why each requests sets session cookie, and this was the reason :P.

PR #4204, opened 2018-05-01
Fix #4167 for PHP 7.2 incompatibility of count() in renderDynamic().

### LINKED ISSUE #4167: PHP 7.2: count(): Parameter must be an array or an object that implements Countable

A lot of tests on PHP 7.2 are failing with

```
count(): Parameter must be an array or an object that implements Countable
```

`count()` throws a warning since PHP 7.2 in case a scalar or non-countable object is passed. We rely on this behavior in many places so code needs to be adjusted.

Need to check how many places are affected and whether its easy to fix. If its hard to fix in a BC way, PHP 7.1 may be the last version that Yii 1.1 runs on.

### More info

- https://wiki.php.net/rfc/counting_non_countables
- http://php.net/manual/en/function.count.php

ThePepster:
@samdark as per your request (https://github.com/yiisoft/yii/issues/4167#issuecomment-381564554)

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4167

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,15 +4,15 @@
 Version 1.1.20 under development
 --------------------------------
 
-- Bug #4179: Fixed MariaDB 10.2 current_timestamp() (yaBliznyk)
-- Bug #4167: Fixed PHP 7.2 incompatibility of "count()" in `CActiveFinder` (softark)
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thanks!

Issue #4198, opened 2018-04-04
PHP 7.2 sqlserver lastInsertId


bchoupartHenji:
Note that only PHP 7 compatibility issues are accepted. For security issues contact maintainers privately.

### What steps will reproduce the problem?
sqlserver 2008 R2
php 7.2.3
driver pdo Microsoft/msphpsql for windows for php 7.2

### What is the expected result?
After saving a new CActiveRecord, the primary key is set 

### What do you get instead?
the primary key is set to '' (empty string)

### Additional info

see [issue](https://github.com/Microsoft/msphpsql/issues/694)

to resolve the problem (i hope  ) : 
in framework/db/schema/mssql/CMssqlSqlsrvPdoAdapter
replace the function
```php
public function lastInsertId($sequence=null)
{
	return parent::lastInsertId();
}
```

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.19
| PHP version      | 7.2.3
| Operating system | windows server 2008 R2

ThePepster:
In the Issue referenced in the previous comment (https://github.com/Microsoft/msphpsql/issues/694) they link to the updated documentation (https://github.com/Microsoft/msphpsql/wiki/Features#lastinsertid).  Based on that I beleive the best course of action is to remove the lastInsertId method from framework/db/schema/mssql/CMssqlSqlsrvPdoAdapter altogether.

I do not have a MSSQL server so I can't test but I would be willing to create a pull request removing this method if all agree.

agusdrs:
I'm going to make a PR with the changes needed for the new version of the SQLSRV driver, maintaining the workaround for the old versions.

ThePepster:
HI,

@agusdrs, @samdark FYI.  That last commit is not compatible with PHP 5.1.  It requires version 5.4 or above.

The offending line is
`intval(explode('.', phpversion('pdo_sqlsrv'))[0])`

PR #4213, opened 2018-06-03
fix PHP 5.1 incomp. in CMssqlSqlsrvPdoAdapter caused by PHP 7.2 fix

vintagesucks:
not a PHP 7.2 fix per se, but the bug was introduced in #4198 (which fixed a PHP 7.2 issue)

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | <a href="https://github.com/yiisoft/yii/issues/4198#issuecomment-394151912">#4198 (comment)</a>


samdark:
Merged. Thanks!

PR #4218, opened 2018-07-03
WIP: Quick fix #4182 - mcrypt dependency

### LINKED ISSUE #4182: the mcrypt is removed in php 7.2 ,so that the /base/CSecurityManager.php can not work.
Note that only PHP 7 compatibility issues are accepted. For security issues contact maintainers privately.

the mcrypt is removed in php 7.2 ,so that the /base/CSecurityManager.php can not work.
the encrypt,decrypt,validateEncryptionKey,etc. use mcrypt module , the mcrypt functions need to be replace by openssl way .
the yii 1.1.x can solve the mcrypt problem in php 7.2.x or above.
### What do you get instead?
I rewrote the encrypt,decrypt and commented out the mcrypt function of the validateEncryptionKey function. 
It just is a temporary solved way,so I need you help to be compatible it.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.19
| PHP version      | 7.2
| Operating system | centos

sergey-dryabzhinsky:
As of php-7.2 mcrypt extension moved to pecl.
But it's required to use CSecurityManager.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4182

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,7 @@
 Version 1.1.20 under development
 --------------------------------
 
-- Bug #4182: CSecurityManager requires mcrypt extension, and for PHP 7.2 we need to install it from pecl (sergey-dryabzhinsky)
+- Bug #4182: CSecurityManager requires mcrypt extension to work, and for PHP 7.2 we need to install it from pecl (sergey-dryabzhinsky)
 - Bug #4162: Adjusted Zend Escaper to be compatible with PHP <5.3 and fixed usage of non-existing Exceptions (cebe)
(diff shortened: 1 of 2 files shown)


sergey-dryabzhinsky:
It's already have checks in `requirements`.
But maybe need to be more verbose if `php` version is `7.2`.

samdark:
Merged. Thanks!

PR #4228, opened 2018-10-17
Fix deprecation error in jquery3.x by not using shortcuts.

kenguest:
remove deprecation errors from framework/web/js/source/jquery.yiiactiveform.js  when using jQuery3.1.1

jquery.yiiactiveform.js has some issues in it that are highlighted when using jquery3.1.1 instead of old jquery1.12.4 library. 

Important as there are security issues in 1.12.4.

This PR fixes them.


kenguest:
It is - I just applied changes to all of the code rather than to just some that had already been made.

kenguest:
I've updated the Pull Request summary - I think I've added the right line for changelog?

samdark:
Merged. Thank you!

PR #4231, opened 2018-11-27
Change to support SQL server functions as default values

advicente:
SQL server functions was being cast to string, inserting string 'getdate' into date/datetime columns.
The expected behavior is to framework set column to NULL and database runs SQL Server function present in default column value.


samdark:
Sorry but bugfixes aren't accepted into 1.1 anymore. Only PHP compatibility fixes and security issues.

Issue #4238, opened 2019-01-22
CJavaScript::quote(null); throws error, incompatible change since Yii 1.1.20


stevoh6:
Note that only PHP 7 compatibility issues are accepted. For security issues contact maintainers privately.

### Background information
Since many applications running on Yii1 are quite old, we are facing one backward incompatible change. 

I'm suggesting add simple check, pull request created. 

### What steps will reproduce the problem?
Using 
`CJavaScript::quote(null); `
throws error.
### What is the expected result?
Do not try escaping/converting null value as it was before https://github.com/yiisoft/yii/commit/1e1c5b7b058eedf3eb2f58914b6be35e965463fc#diff-34b1d76ab6620e419eaef5799cff7103. 
### What do you get instead?
`Exception
String to be escaped was not valid UTF-8 or could not be converted:
/var/www/app/protected/vendor/yiisoft/yii/framework/vendors/zend-escaper/Escaper.php(321)`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.20
| PHP version      | 7.2
| Operating system | Alpine

rob006:
> Using
> `CJavaScript::quote(null); `
> throws error.

This looks like a bug in your app. Why would you want to quote `null`?

samdark:
It was never type-hinted to accept null so I'd like to hear the reason as well.

stevoh6:
Hi, I fully understand, that this is unusual usage this function, I was take over couple old projects and it was pretty surprise for me how it was implemented in our projects, but I get the idea why. Of course it can be rewrite and Im already make some adjustment to avoid this "behavior", but it would be to nice don't change old behavior. So I decided open pull request, maybe someone have similar "problem" :-)

One of real (not my, not really good) example: from controller you get some settings values (stored in db), you pass this whole array to view and you trying put those values into js variables (for example to fill input-boxes) and you would like to quoted them regardless their value.

ddziaduch:
Even the method signature shows that you should pass a string: `public static string quote(string $js, boolean $forUrl=false)`

ddziaduch:
I can fix this 🙂

PR #4239, opened 2019-01-22
Update CJavaScript::quote to accept null value. Backward compatibility change.

stevoh6:
Add backward compatibility to do not escape null value. As it was before Yii 1.1.20

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4238

### DIFF
--- a/framework/web/helpers/CJavaScript.php
+++ b/framework/web/helpers/CJavaScript.php
@@ -21,7 +21,7 @@ class CJavaScript
 	 * Quotes a javascript string.
 	 * After processing, the string can be safely enclosed within a pair of
 	 * quotation marks and serve as a javascript string.
-	 * @param null|string $js string to be quoted
+	 * @param string $js string to be quoted
 	 * @param boolean $forUrl whether this string is used as a URL
 	 * @return string the quoted string
 	 */
--- a/framework/web/helpers/CJavaScript.php
+++ b/framework/web/helpers/CJavaScript.php
@@ -25,18 +25,19 @@ class CJavaScript
 	 * @param boolean $forUrl whether this string is used as a URL
 	 * @return null|string the quoted string
 	 */
-	public static function quote($js,$forUrl=false)
+        public static function quote($js,$forUrl=false)
 	{
-		if ($js === null) { // backward compability before Yii 1.1.20
-			return $js;
-		}
+		if($js===null)
+		{
+			return null;
+                }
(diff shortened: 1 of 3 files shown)


stevoh6:
> If we want it we should do `(string)$js` type casting instead of checking for null.

Done.

framework/web/helpers/CJavaScript.php
```diff
@@ -21,18 +21,21 @@ class CJavaScript
...
+        public static function quote($js,$forUrl=false)
 	{
-        Yii::import('system.vendors.zend-escaper.Escaper');
-        $escaper=new Escaper(Yii::app()->charset);
-        if($forUrl)
-            return $escaper->escapeUrl($js);
-        else
-            return $escaper->escapeJs($js);
+		if($js===null)
+			return null;
```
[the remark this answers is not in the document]
stevoh6: Done.

samdark:
Would you please apply test from #4241?

framework/web/helpers/CJavaScript.php
```diff
@@ -21,18 +21,20 @@ class CJavaScript
 	 * Quotes a javascript string.
 	 * After processing, the string can be safely enclosed within a pair of
 	 * quotation marks and serve as a javascript string.
-	 * @param string $js string to be quoted
+	 * @param null|string $js string to be quoted
```
rob006: I would leave this as it was before - `null` is not a special case here, right now methods supports everything that can be cast to string.

stevoh6:
> Would you please apply test from #4241?

Done.

tests/framework/web/helpers/CJavaScriptTest.php
```diff
@@ -42,4 +42,12 @@ public function testQuoteForUrl()
         $output=CJavaScript::quote($input,true);
         $this->assertEquals('test%20%E2%80%A8%0Atest%20%E2%80%A9',$output);
     }
+	
+    public function testQuoteWithNull()
+    {
+        $input=null;
+        $output=CJavaScript::quote($input);
+        $this->assertEquals('',$output);
```
rob006:
```suggestion
        $this->assertSame('',$output);
```

samdark:
Merged. Thank you!

PR #4240, opened 2019-01-23
YiiBase::autoload should not try to include() missing files

sanmai:
Make sure autoloading does not try to load files that do not exist. Else it'll cause intermittent warnings.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | n/a

### DIFF
--- a/framework/YiiBase.php
+++ b/framework/YiiBase.php
@@ -460,7 +460,12 @@ class YiiBase
 					}
 				}
 				else
-					include($className.'.php');
+				{
+					if(is_file($className.'.php'))
+						include($className.'.php');
+					else
+						return false;
+				}
 			}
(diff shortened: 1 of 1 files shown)


rob006:
You can set `Yii::$enableIncludePath` to `false` if you want to disable this behavior.

sanmai:
That's a nice workaround! 

But still, a proper class loader should be returning `false` instead of warnings. What do you think?

rob006:
This is not a workaround, this is how you should configure Yii autoloader if you don't want warnings/exceptions. Current behavior is kept for BC and performance reasons, it cannot be changed.

PR #4241, opened 2019-01-28
Handle null in `CJavaScript::quote`

ddziaduch:
Ref #4238

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4238

### DIFF
--- a/framework/web/helpers/CJavaScript.php
+++ b/framework/web/helpers/CJavaScript.php
@@ -21,12 +21,14 @@ class CJavaScript
 	 * Quotes a javascript string.
 	 * After processing, the string can be safely enclosed within a pair of
 	 * quotation marks and serve as a javascript string.
-	 * @param string $js string to be quoted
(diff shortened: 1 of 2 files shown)


framework/web/helpers/CJavaScript.php
```diff
@@ -21,12 +21,14 @@ class CJavaScript
 	 * Quotes a javascript string.
 	 * After processing, the string can be safely enclosed within a pair of
 	 * quotation marks and serve as a javascript string.
-	 * @param string $js string to be quoted
+	 * @param string|null $js string to be quoted
```
rob006: I would leave this as it was before - `null` is not a special case here, right now methods support everything that can be cast to string.

samdark:
Thanks for pull requesting. Your test is in.

PR #4255, opened 2019-04-29
don't call afterDelete when delete failed

DrDeath72:
when catch delete exception (fk, etc...) and return rowsCount=0 afterDelete still calling

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no


samdark:
Can not accept the fix since:

> Note that only PHP 7 compatibility fixes are accepted

Issue #4256, opened 2019-04-29
PHP 7 compatibility issues: PHP4 style constructor and create_function()


kenguest:
gii/components/Pear/Text/Diff3.php:
Use of deprecated PHP 4 style class constructor is not supported in PHP7+

vendors/markdown/markdown.php:
create_function() is deprecated since PHP7.2 - should be using an anonymous function instead.

samdark:
Constructor part solved.

kenguest:
I've reinstated use of create_function in markdown.php so to keep compatibility with 5.1.x as requested.

Looking at the changelog it seems others replaced create_function calls with anonymous functions elsewhere, facilitating migration to a maintained version of PHP. 

This is why I'm making these changes in the first place....

samdark:
@kenguest no. https://github.com/yiisoft/yii/commit/68e5e86a11162a54673d83f8d0a5939b621d8040

PR #4257, opened 2019-04-29
#4256: PHP 7 compatibility issues.

kenguest:
Bug #4256: PHP 7 compatibility issues: PHP4 style constructor in Pear/Text/Diff3.php and create_function() in markdown/markdown.php (kenguest)

### DIFF
--- a/framework/vendors/markdown/markdown.php
+++ b/framework/vendors/markdown/markdown.php
@@ -1455,9 +1455,9 @@ class Markdown_Parser {
 	# regular expression.
 	#
 		if (function_exists($this->utf8_strlen)) return;
-		$this->utf8_strlen = function ($text) use ($m) {
-			return preg_match_all("/[\\x00-\\xBF]|[\\xC0-\\xFF][\\x80-\\xBF]*/", $text, $m);
-		};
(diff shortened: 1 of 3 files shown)


framework/vendors/markdown/markdown.php
```diff
@@ -1455,9 +1455,9 @@ public function _initDetab() {
 	# regular expression.
 	#
 		if (function_exists($this->utf8_strlen)) return;
-		$this->utf8_strlen = create_function('$text', 'return preg_match_all(
-			"/[\\\\x00-\\\\xBF]|[\\\\xC0-\\\\xFF][\\\\x80-\\\\xBF]*/",
-			$text, $m);');
+		$this->utf8_strlen = function ($text) use ($m) {
```
samdark: Can't be done like that because `"php": ">=5.1.0"`. `create_function` should be left as fallback.

kenguest: I understand, but what are the chances of people still using php 5.1 though? ;-)

samdark: [About 7/100](https://w3techs.com/technologies/details/pl-php/5/all).

samdark:
Merged. Thank you!

Issue #4261, opened 2019-05-23
PHP 7.3 CUrlValidator with validateIDN option: INTL_IDNA_VARIANT_2003 is deprecated


npushkov:
Note that only PHP 7 compatibility issues are accepted. For security issues contact maintainers privately.

### What steps will reproduce the problem?

Validate any url via CUrlValidator with `validateIDN` option set to `true`

### What is the expected result?

Url is validated without PHP errors

### What do you get instead?

PHP Error [500]
idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.21
| PHP version      | 7.3
| Operating system |

machour:
@npushkov should be fixed by #4262, could you give it a try?

npushkov:
@machour I have tried it, #4262 fixes this issue. Thank you!

machour:
Thank you, let's keep this issue open until the PR gets merged.

samdark:
Its merged. @machour thank you for looking into it.

PR #4262, opened 2019-05-26
Work around INTL_IDNA_VARIANT_2003 deprecation

machour:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | #4261

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 7.4snapshot
   - 7.3
   - 7.2
   - 7.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 php:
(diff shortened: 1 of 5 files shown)


machour:
So full story here: PHP deprecated INTL_IDNA_VARIANT_2003 in 7.2, but kept it as default value for the parameter.

The default value will be changed in 7.4. And INTL_IDNA_VARIANT_2003 removed in PHP 8.

Relevant RFC here: https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003

-- 

Had to apply a fix on top of htmlpurifier 4.10.0 which is merge in master and prevent usage of undefined constants. Not sure when 4.10.1 will be released.

PR #4266, opened 2019-07-08
Fix limit = NULL

alexpts:
I have block `SELECT 0` on $limit=null.

Actual: SELECT ... LIMIT 0
Expected: SELECT ...

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no


samdark:
Won't be applied since it's neither security fix nor compatibility fix.

Issue #4269, opened 2019-10-04
bindValues() causing PHP 7.2 compatibility error from count()


krunder:
Note that only PHP 7 compatibility issues are accepted. For security issues contact maintainers privately.

### What steps will reproduce the problem?

Passing null through the $values parameter on the bindValues() method in CdbCommandBuilder.php.

### What is the expected result?

The count function return 0 as would usually be the case with versions of PHP below 7.2.

### What do you get instead?

PHP fatal error complaining that the value passed to count() should be an array or implement Countable.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.21
| PHP version      | 7.2 & 7.3
| Operating system | Windows & Linux

rob006:
> Passing null through the $values parameter on the bindValues() method in CdbCommandBuilder.php.

This method expects array. In which case `null` will be passed?

krunder:
Apologies for the late reply on this. It looks to be that null is generally passed into the bind values if non are specified. I experienced this with one of the projects I am working on at work.

samdark:
@krunder why would bindValues() be called if nothing is being bound?

PR #4270, opened 2019-10-04
4269 bind values null

krunder:
PHP 7.2 compatibility issues: bindValues() not defaulting to array when null causing error with count().

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4269


thekonz:
The method signature does not declare the type to be array, the docblock does.

> public function bindValues($command, $values)

So it is possible to enter any value.

samdark:
@thekonz I tend to agree with @b1rdex about it. While it's possible to pass anything in PHP less than 7, it's documented that the method is accepting array only. Passing `null` definitely looks like incorrect usage from the application side.

Waiting for an input from @krunder, https://github.com/yiisoft/yii/issues/4269#issuecomment-553776236.

samdark:
Decided not to merge the pull request. Thanks for effort, @krunder but null shouldn't be passed in the first place.

Issue #4272, opened 2019-10-31
PHP Error[2]: include(ParagonIE_Sodium_Compat.php): failed to open stream: No such file or directory


ursuleacv:
When installing `paragonie/sodium_compat` I'm getting the following error:

**PHP Error[2]: include(ParagonIE_Sodium_Compat.php): failed to open stream: No such file or directory**

The class ParagonIE_Sodium_Compat cannot be found in **cli** mode, works fine in php-fpm.

### What steps will reproduce the problem?
- Install Yii framework 1.x
- cd protected
- run composer require paragonie/sodium_compat
- create a command for example TestCommand
- run php cron.php test

### What is the expected result?
work without any errors

### What do you get instead?

```
PHP Error[2]: include(ParagonIE_Sodium_Compat.php): failed to open stream: No such file or directory
    in file C:\localhost\framework\YiiBase.php at line 463
#0 C:\localhost\framework\YiiBase.php(463): include()
#1 unknown(0): autoload()
#2 C:\localhost\protected\vendor\paragonie\sodium_compat\lib\constants.php(8): spl_autoload_call()
#3 C:\localhost\protected\vendor\paragonie\sodium_compat\lib\sodium_compat.php(825): require_once()
#4 C:\localhost\protected\vendor\paragonie\sodium_compat\autoload.php(45): require_once()
#5 C:\localhost\protected\vendor\composer\autoload_real.php(70): require()
#6 C:\localhost\protected\vendor\composer\autoload_real.php(60): composerRequire50966560348a0ac36e4b9ee4405e5f09()
#7 C:\localhost\protected\vendor\autoload.php(7): getLoader()
#8 C:\localhost\protected\commands\TestCommand.php(7): require_once()
#9 C:\localhost\framework\console\CConsoleCommandRunner.php(166): require_once()
#10 C:\localhost\framework\console\CConsoleCommandRunner.php(67): CConsoleCommandRunner->createCommand()
#11 C:\localhost\framework\console\CConsoleApplication.php(92): CConsoleCommandRunner->run()
#12 C:\localhost\framework\base\CApplication.php(185): CConsoleApplication->processRequest()
#13 C:\localhost\cron.php(19): CConsoleApplication->run()
```
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.21
| PHP version      | 7.0.33
| Operating system | Windows 10 or Centos 6.6

rob006:
[...]

PR #4274, opened 2019-11-07
Trim namespace separators before checking string

paragonie-scott:
See https://github.com/paragonie/sodium_compat/issues/105

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,7 @@ Version 1.1.22 under development
 
 - Bug #4256: PHP 7 compatibility issues: PHP4 style constructor in Pear/Text/Diff3.php (kenguest)
 - Bug #4261: PHP 7.2 compatibility issues: INTL_IDNA_VARIANT_2003 has been depreacated (machour)
+- Chg #4274: PHP 7.2 compatibility issues: Autoloader failing with third-party libraries using PSR-0 namespaces with a `\` prefix.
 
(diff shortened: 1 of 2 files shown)


samdark:
Look good. Would you please add a line for cangelog file?

samdark:
Note to myself: check why tests fail before merge. It should not be related though.

rob006:
It does not make sense for me. Autoloader should always get `$className` without leading backslash. 

I'm not sure how this could fix https://github.com/paragonie/sodium_compat/issues/105 either (in stacktrace you can see that error is thrown inside of `if(strpos($className,'\\')===false)` branch - this PR will not change that, even if `$className` would contain leading backslash).

paragonie-scott:
https://github.com/yiisoft/yii/blob/83a86bca722e438f2e402d1e1cd7a723d41da12a/framework/YiiBase.php#L462-L463

Right, it looks like `self::$enableIncludePath` is the culprit instead. I wonder if there's any way to coerce their other dependencies to load sodium_compat first before hitting this autoloader?

samdark:
So it seems the pull request would not help :( Closing. Thanks for trying and what you contributed to Yii before.

samdark:
See https://github.com/yiisoft/yii/issues/4272

Issue #4273, opened 2019-11-06
setcookie $options to allow setting of SameSite flag


thekonz:
Google Chrome will soon (January 2020) basically block third-party cookies, if they do not have the SameSite flag set. See [this](https://www.chromestatus.com/feature/5088147346030592) and [that](https://www.chromestatus.com/feature/5633521622188032) for more information on the Chrome side of things.

PHP 7.3 changed `setcookie` to receive an `$options` parameter, which allows to set the SameSite flag.

**Will Yii 1.1 get an update to support the $options parameter?** Does this fall in the category of compatibility updates? 

### What steps will reproduce the problem?

Set a cookie while being a third party (ex. run your website in an iframe) in Chrome.

### What is the expected result?

The cookie should be set.

### What do you get instead?

From January 2020 on, Chrome will block this cookie.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.21
| PHP version      | 7.3

samdark:
Yes. Look like exactly that category. Do you want to prepare a pull request?

thekonz:
Sure, will do later.

thekonz:
Well, the PR is there, but travis complains about some mysql server not being available.

PR #4275, opened 2019-11-08
#4273 add setcookie $options for PHP 7.3 compatibility

thekonz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4273

### DIFF
--- a/framework/web/CHttpCookie.php
+++ b/framework/web/CHttpCookie.php
@@ -48,6 +48,11 @@ class CHttpCookie extends CComponent
 	 * Note, this property is only effective for PHP 5.2.0 or above.
 	 */
 	public $httpOnly=false;
+    /**
+     * @var array Additional setcookie options like "SameSite".
+     * This property only works for PHP 7.3.0 or above.
+     */
+	public $options=array();
 
 	/**
 	 * Constructor.
--- a/framework/web/CHttpCookie.php
+++ b/framework/web/CHttpCookie.php
@@ -18,10 +18,22 @@
 class CHttpCookie extends CComponent
 {
 	/**
-	 * Possible values for attribute "sameSite".
-	 * @see https://www.owasp.org/index.php/SameSite
+	 * SameSite policy Lax will prevent the cookie from being sent by the browser in all cross-site browsing context
+	 * during CSRF-prone request methods (e.g. POST, PUT, PATCH etc).
+	 * E.g. a POST request from https://otherdomain.com to https://yourdomain.com will not include the cookie, however a GET request will.
+	 * When a user follows a link from https://otherdomain.com to https://yourdomain.com it will include the cookie
+	 * @see $sameSite
+	 * @since 1.1.22
 	 */
 	const SAME_SITE_LAX='Lax';
+	/**
+	 * SameSite policy Strict will prevent the cookie from being sent by the browser in all cross-site browsing context
+	 * regardless of the request method and even when following a regular link.
+	 * E.g. a GET request from https://otherdomain.com to https://yourdomain.com or a user following a link from
+	 * https://otherdomain.com to https://yourdomain.com will not include the cookie.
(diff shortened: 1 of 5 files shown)


thekonz:
@samdark Any idea why this is failing? My test works on its own, but travis just cannot connect to the mysql server.

My6UoT9:
According to https://github.com/travis-ci/travis-ci/issues/6842#issuecomment-502259411
> https://travis-ci.community/t/prs-started-failing-on-6-12/3707/2
>
>We are moving to Xenial as the default, and we do not start DB by default.
>
>https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment#1-services-support
>
>You need to explicitly instruct that it is needed.
>
>services:
>  - mysql

https://github.com/yiisoft/yii/blob/master/.travis.yml
So maybe just add mysql to the services.
 This is strange since, the tests should be failing since a while - if this the reason?

My6UoT9:
Seems mysql works now, but it wants psql now..
https://docs.travis-ci.com/user/database-setup/#postgresql
>services:
>  - postgresql

My6UoT9:
Travis using xenial (`Build dist: xenial`)  instead of trusty.
Xenial seems to not support older php versions, trusty should.

I would try to revert the changes to services, and set dist to trusty instead of xenial (which is default).

https://docs.travis-ci.com/user/reference/trusty/
Would need this key in main (as seen in this example: https://docs.travis-ci.com/user/reference/trusty/#hhvm).
`dist: trusty`

thekonz:
@My6UoT9 thank you :D

framework/web/CHttpCookie.php
```diff
@@ -48,6 +48,11 @@ class CHttpCookie extends CComponent
 	 * Note, this property is only effective for PHP 5.2.0 or above.
 	 */
 	public $httpOnly=false;
+    /**
+     * @var array Additional setcookie options like "SameSite".
+     * This property only works for PHP 7.3.0 or above.
+     */
+	public $options=array();
```
samdark: It seems there are no additional options other than SameSite. How about handing it the same way it's done in Yii 2: https://github.com/yiisoft/yii2/blob/master/framework/web/Cookie.php#L80 and Yii 3: https://github.com/yiisoft/yii-web/blob/master/src/Cookie.php#L79

framework/web/CHttpCookie.php
```diff
@@ -48,6 +48,11 @@ class CHttpCookie extends CComponent
 	 * Note, this property is only effective for PHP 5.2.0 or above.
 	 */
 	public $httpOnly=false;
+    /**
+     * @var array Additional setcookie options like "SameSite".
```
samdark: Formatting seems to be off. Not a huge issue though. Can fix it while merging.

framework/web/CHttpCookie.php
thekonz: > there are no additional options other than SameSite

... yet

I think that we want the code to be flexible for the future and prevent more pull requests for every other flag that browser vendors want to add.

framework/web/CHttpCookie.php
thekonz: Fixed it.

framework/web/CHttpCookie.php
samdark: That complicates the syntax. Additional properties could be added as needed.

thekonz: > That complicates the syntax

That is a bold claim, but hey... have it your way.
I changed it.

samdark:
Merged. Thank you!

tomfotherby:
I propose to also update `setCookieParams`. I raised a PR: https://github.com/yiisoft/yii/pull/4304

PR #4285, opened 2020-01-07
Fix #4256: Fixed deprecated usage of create_function() in markdown.php

### LINKED ISSUE #4256: PHP 7 compatibility issues: PHP4 style constructor and create_function()

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4256

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,8 +4,9 @@
 Version 1.1.22 under development
 --------------------------------
 
-- Bug #4256: PHP 7 compatibility issues: PHP4 style constructor in Pear/Text/Diff3.php (kenguest)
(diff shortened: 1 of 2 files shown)


rob006:
I don't think it will work on PHP 5.2 - PHP will not be able to parse this file.

b1rdex:
We can solve it through introducing private method and passing it as callback. 

Отправлено с iPhone

samdark:
Changed implementation.

PR #4286, opened 2020-01-08
Update phpdoc in CHtml

rob006:
`CHtml::modelName()` (used by all these functions) supports also `string` as a `$model`.


samdark:
Thanks!

PR #4292, opened 2020-03-05
Fix #4291: The scheme(protocol) is deleted when validateIDN is enabled after validation

### LINKED ISSUE #4291: The scheme(protocol) is deleted when validateIDN is enabled after validation
1. Add a validation rule to the model.
```php
<?php
class ModelName extends CActiveRecord{
    ... other methods
    public function rules(){
        return [
            ... other rules
            ['url_field', 'url', 'validateIDN'=>true]
        ];
    }
}
```
2. Enter data
```php
$modelName = new ModelName;
$model->url_field = 'https://example.com/full/path';
```
3. Validation
```php
$modelName->validate();
```

```php
echo $model->url_field; # https://example.com/full/path
```

### What do you get instead?
```php
echo $model->url_field; # example.com/full/path
```

### Additional info
Missing line concatenation.
https://github.com/yiisoft/yii/blob/master/framework/validators/CUrlValidator.php#L213

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.22 (commit 83a86bca722e438f2e402d1e1cd7a723d41da12a)
| PHP version      | 7.2
| Operating system | All

Argevollen:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4291

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,7 @@
 Version 1.1.23 under development
 --------------------------------
 
-no changes yet.
+- Bug #4291: The scheme (protocol) is deleted when validateIDN is enabled after validation (Argevollen)
 
 Version 1.1.22 January 16, 2020
 -------------------------------
--- a/framework/validators/CUrlValidator.php
+++ b/framework/validators/CUrlValidator.php
@@ -210,7 +210,7 @@ if(jQuery.trim(value)!='') {
 				}
 				else
 				{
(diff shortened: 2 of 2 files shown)


samdark:
Thank you!

Issue #4294, opened 2020-04-01
Postgresql 12 issue in CPgsqlSchema.php


d4rkstar:
### What steps will reproduce the problem?
Use postgresql 12

### What is the expected result?
Application should work without problems

### What do you get instead?
SQLSTATE[42703]: Undefined column: 7 ERROR: column c.consrc does not exist

### Additional info
Steps to reproduce or a small repository showing the problem: PostgreSQL 12 removes the pg_constraint.consrc column and this breaks CPgsqlSchema.php

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.22
| PHP version      | 7.2.27
| Operating system | Debian 10

samdark:
Is there a simple way to fix it? If there isn't I'm afraid we can't have it since Yii 1.1 should only receive security fixes and PHP 7 compatibility fixes.

d4rkstar:
Hello @samdark . Yes, i've developed a fix and i'll be glad to send a patch, but don't know how to do extensive tests.
Will send it later today.

d4rkstar:
Take look at #4296

romansh:
Faced the same issue. Used bio's commits to change yii code and waiting for 1.2.3.

DrDeath72:
PHP notice
Undefined index: column_def_value
PostgreSQL 10.14 (Ubuntu 10.14-0ubuntu0.18.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0, 64-bit

marcovtwout:
@d4rkstar @DrDeath72 Could you please test the new solution in https://github.com/yiisoft/yii/pull/4323 and confirm on which PostgreSQL versions it works as expected?

d4rkstar:
@marcovtwout Yes, i can test. Will let you know.

Regards

d4rkstar:
@marcovtwout #4323 seems to work for me. No issue so far

marcovtwout:
@d4rkstar Thanks - and I assume that is on v12?

d4rkstar:
Yes

marcovtwout:
Fixed with https://github.com/yiisoft/yii/pull/4323

PR #4296, opened 2020-04-02
In Postgresql 12 pg_attrdev.adsrc column is removed (https://www.post…

d4rkstar:
…gresql.org/docs/12/release-12.html): This column has been deprecated for a long time, because it did not update in response to other catalog changes (such as column renamings). The recommended way to get a text version of a default-value expression from pg_attrdef is pg_get_expr(adbin, adrelid).

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | don't know
| Tests pass?   | ✔️
| Fixed issues  | #4294

### DIFF
--- a/framework/db/schema/pgsql/CPgsqlSchema.php
+++ b/framework/db/schema/pgsql/CPgsqlSchema.php
@@ -166,7 +166,7 @@ class CPgsqlSchema extends CDbSchema
 	protected function findColumns($table)
 	{
 		$sql=<<<EOD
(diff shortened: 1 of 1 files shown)


samdark:
Need testing on lower versions of PostgreSQL.

samdark:
Solved by another pull request. Thanks for contributing anyway.

d4rkstar:
No problem @samdark . No time at all :(:(

PR #4297, opened 2020-04-07
Add PostgreSQL 12 support, rel #4294

bio:
Add PostgreSQL 12 support.

From PostgreSQL 12 [changelog](https://www.postgresql.org/docs/release/12.0/):
- Remove obsolete pg_constraint.consrc column (Peter Eisentraut)
- Remove obsolete pg_attrdef.adsrc column (Peter Eisentraut)

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4294

### DIFF
--- a/framework/db/schema/pgsql/CPgsqlSchema.php
+++ b/framework/db/schema/pgsql/CPgsqlSchema.php
@@ -165,13 +165,13 @@ class CPgsqlSchema extends CDbSchema
 	 */
 	protected function findColumns($table)
 	{
-        $serverVersion = $this->getDbConnection()->getServerVersion();
-        $columnDefValue = version_compare($serverVersion, '12.0', '<')
-            ? 'd.adsrc' : 'CAST(pg_get_expr(d.adbin, d.adrelid) AS varchar)';
+		$serverVersion = $this->getDbConnection()->getServerVersion();
+		$columnDefValue = version_compare($serverVersion, '12.0', '<')
+			? 'd.adsrc' : 'CAST(pg_get_expr(d.adbin, d.adrelid) AS varchar)';
 
 		$sql=<<<EOD
 SELECT a.attname, LOWER(format_type(a.atttypid, a.atttypmod)) AS type,
-    {$columnDefValue} AS column_def_value, a.attnotnull, a.atthasdef,
+	{$columnDefValue} AS column_def_value, a.attnotnull, a.atthasdef,
 	pg_catalog.col_description(a.attrelid, a.attnum) AS comment
 FROM pg_attribute a LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
 WHERE a.attnum > 0 AND NOT a.attisdropped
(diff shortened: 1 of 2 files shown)


samdark:
Need testing on lower PostgreSQL versions.

bio:
@samdark yes, it's compatible. I didn't change the previous implementation, just added postgresql 12+ support.

samdark:
Have you seen https://github.com/yiisoft/yii/pull/4296?

bio:
I saw #4296 but I hadn't tested it.

samdark:
Thanks!

framework/db/schema/pgsql/CPgsqlSchema.php
```diff
@@ -186,7 +191,7 @@ protected function findColumns($table)
 			$c=$this->createColumn($column);
 			$table->columns[$c->name]=$c;
 
-			if(stripos($column['adsrc'],'nextval')===0 && preg_match('/nextval\([^\']*\'([^\']+)\'[^\)]*\)/i',$column['adsrc'],$matches))
+			if(stripos($column['column_def_value'],'nextval')===0 && preg_match('/nextval\([^\']*\'([^\']+)\'[^\)]*\)/i',$column['column_def_value'],$matches))
```
cebe, line 194: for full BC, the name `adsrc` should be kept using an alias in the query above.

bio: hm, if someone has extended the CPgsqlSchema findColumns method, changes in the original method won't affect their new code.

cebe: `protected function createColumn($column)` depends on the name, so if someone extended that one, it might break.

framework/db/schema/pgsql/CPgsqlSchema.php
```diff
@@ -165,8 +165,13 @@ protected function resolveTableNames($table,$name)
 	 */
 	protected function findColumns($table)
 	{
+		$serverVersion = $this->getDbConnection()->getServerVersion();
+		$columnDefValue = version_compare($serverVersion, '12.0', '<')
+			? 'd.adsrc' : 'CAST(pg_get_expr(d.adbin, d.adrelid) AS varchar)';
```
marcovtwout, line 170: @bio Can you explain why the CAST is needed?

bio: @marcovtwout just to maek sure we always have string. `adbin` has type `pg_node_tree` and `adrelid` has type `oid` (https://www.postgresql.org/docs/12/catalog-pg-attrdef.html)

marcovtwout: Correct me if I'm wrong, but I don't think casting needs to apply here. The types you mension are the parameter types `pg_get_expr()`, which has a return type of `text` (https://www.postgresql.org/docs/12/functions-info.html#FUNCTIONS-INFO-CATALOG-TABLE).

bio: @marcovtwout yeah, pg_get_expr returns text so no need to cast to varchar.

marcovtwout:
Both issues corrected in https://github.com/yiisoft/yii/pull/4323

PR #4304, opened 2020-07-31
add session_set_cookie_params samesite option for PHP 7.3 compatibility

tomfotherby:
In PHP 7.3 the [session_set_cookie_params](https://www.php.net/manual/en/function.session-set-cookie-params.php) function had the following change:

> 7.3.0 | An alternative signature supporting an options array has been added. **This signature supports also setting of the `SameSite` cookie attribute**.

This PR bring Yii up to date for PHP 7.3. This PR is a follow-up to https://github.com/yiisoft/yii/pull/4275 (_add setcookie $options for PHP 7.3 compatibility_)

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/framework/web/CHttpSession.php
+++ b/framework/web/CHttpSession.php
@@ -241,7 +241,7 @@ class CHttpSession extends CApplicationComponent implements IteratorAggregate,Ar
 	 * The effect of this method only lasts for the duration of the script.
 	 * Call this method before the session starts.
 	 * @param array $value cookie parameters, valid keys include: lifetime, path,
-	 * domain, secure, httponly. Note that httponly is all lowercase.
+	 * domain, secure, httponly, samesite. Note that httponly and samesite is all lowercase.
 	 * @see http://us2.php.net/manual/en/function.session-set-cookie-params.php
 	 */
 	public function setCookieParams($value)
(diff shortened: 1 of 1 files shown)


samdark:
It won't work as it is. Here's how it can be done: https://github.com/yiisoft/yii2/commit/975937e531b488864047f306c047a2240d3bab59

tomfotherby:
> It won't work as it is. Here's how it can be done: [yiisoft/yii2@975937e](https://github.com/yiisoft/yii2/commit/975937e531b488864047f306c047a2240d3bab59)

Thanks @samdark . I took another stab at it.

tomfotherby:
Closing because the build fails on "PHP 5.3-ubuntu-latest" due to:

> PHP Parse error:  syntax error, unexpected '[', expecting ')' in /home/runner/work/yii/yii/framework/web/CHttpSession.php on line 255

Since Yii 1 needs to support older PHP versions, there's not a good solution.

rob006:
> Since Yii 1 needs to support older PHP versions, there's not a good solution.

You can use old array syntax - `array()` works perfectly fine in PHP 8.

tomfotherby:
Re-opened, after changing `[]` to `array()`.

PR #4306, opened 2020-08-05
PHP 8 support

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -


framework/console/CConsoleCommand.php
```diff
@@ -129,7 +129,13 @@ public function run($args)
 			$name=$param->getName();
 			if(isset($options[$name]))
 			{
-				if($param->isArray())
+				if(version_compare(PHP_VERSION,'8.0','>=')) {
+					$isArray=$param->getType() && $param->getType()->getName()==='array';
+				} else {
+					$isArray = $param->isArray();
+				}
```
marcovtwout, line 136: Code formatting could be improved a little bit.

framework/web/actions/CAction.php
```diff
@@ -94,7 +94,13 @@ protected function runWithParamsInternal($object, $method, $params)
 			$name=$param->getName();
 			if(isset($params[$name]))
 			{
-				if($param->isArray())
+				if(version_compare(PHP_VERSION,'8.0','>=')) {
+					$isArray=$param->getType() && $param->getType()->getName()==='array';
+				} else {
+					$isArray=$param->isArray();
+                }
```
marcovtwout, line 101: Code formatting could be improved a little bit.

framework/utils/CPropertyValue.php
```diff
@@ -53,7 +53,7 @@ class CPropertyValue
 	public static function ensureBoolean($value)
 	{
 		if (is_string($value))
-			return !strcasecmp($value,'true') || $value!=0;
+			return !strcasecmp($value,'true') || ($value!=0 && $value!=='' && is_numeric($value));
```
marcovtwout, line 56: @samdark Do you have examples of `$value` that are fixed with this change? Just want to make sure this change is correct.

samdark: @marcovtwout revert this like in master and run tests on PHP 8. That's covered with tests. It is about https://wiki.php.net/rfc/string_to_number_comparison

My6UoT9:
fyi use `\version_compare` instead of `version_compare`
The later looks up the function in current namespace, and only after finds the correct one.
image

see page 63/64 https://www.slideshare.net/nikita_ppv/php-performance-trivia

PR #4311, opened 2020-08-19
port code from yii2 handling memcached issue, related to use of unix timestamp

nikolasr200:
There is an issue with Yii1 and memcached that has already been addressed in yii2. The issue is related to internal memcached clock and how the use of unix timestamp in small expiration examples, doesn't expire the cached items as expected.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️


samdark (approved):
Please add a line for CHANGELOG.

samdark:
Thanks!

PR #4314, opened 2020-09-01
Improve error/exception details when loading fixtures

marcovtwout:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | https://github.com/tremani/yii/pull/1

Changes in this PR:
- Chg: Exceptions thrown while loading fixtures now contain details about the error location (BBoom)
- Chg: Missing fixture files now throw exceptions (BBoom)

(@samdark I know that Yii 1 is officially only accepting compatibility fixes, but I figured it could still be useful for the community. This PR comes from our company private repo where it has been integrated for years. We plan to continue Yii 1 support, if that is welcome I'd rather we contribute here instead of only in our private repo. :) Thank you for considering our PR, feel free to reject if it doesn't fit.)


samdark:
@marcovtwout we do not plan to actively invest time into 1.1 after next release that adds PHP 8 support (and it was part of a deal "get it done and we'll donate to the Yii fund"). Originally it was planned to [end support by the end of the year](https://www.yiiframework.com/release-cycle).

I know that you're around since the very beginning of the project so if you want to take over 1.1 maintenance, we can discuss that.

marcovtwout:
@samdark I'd be happy to continue contributing to maintainance and compatibility support. :) I'll send you a DM so we can discuss further.

CHANGELOG
```diff
@@ -11,6 +11,8 @@ Version 1.1.23 under development
 - Bug #4306: Add PHP 8 support (samdark)
 - Enh #4308: PHP 7.3 compatibility: Add `samesite` as a session cookie option (tomfotherby)
 - Bug #4310: Items on memcache won't expire due to memcache difference in internal clock (nikolasr200)
+- Chg: Exceptions thrown while loading fixtures now contain details about the error location (BBoom)
+- Chg: Missing fixture files now throw exceptions (BBoom)
```
samdark, line 15:
```suggestion
- Chg #4314: Exceptions thrown while loading fixtures now contain details about the error location (BBoom)
- Chg #4314: Missing fixture files now throw exceptions (BBoom)
```

samdark: "Missing fixture files now throw exceptions" could be a problem. As far as I understand, previously they weren't loaded at all, right?

marcovtwout: That is correct, previously they were discarded silently. I would consider that a bug, and regard this change as a bugfix instead of "normal behavior" and a BC-break. What do you think?

samdark: OK. That makes sense,.

b1rdex:
Hi @marcovtwout,

As you're now a member contributor, you may be interested in adopting some features/bug fixes from my fork: https://github.com/yiisoft/yii/compare/master...b1rdex:master if you have time.

Most of the commits are atomic and can be applied w/o conflicts. But there are also some harmful things like dropping a bunch of files or backwards-incompatible things that should be skipped, be careful.

PS. for anyone who's going to look at it: it's better to do it commit-by-commit rather than a complete diff.

marcovtwout:
@b1rdex Note that even although we may make some minor changes, focus is still on compatibility and security issues. But we can at least take a look at it. In that case, could you create a new issue and name some of the relevant features/fixes/commits that you think are suitable to be merged?

samdark (approved):

PR #4316, opened 2020-09-14
Set default value to None on sameSite

nikolasr200:
According to PHP manual https://www.php.net/manual/en/function.session-set-cookie-params.php
"If any of the allowed options are not given, their default values are the same as the default values of the explicit parameters. If the samesite element is omitted, no SameSite cookie attribute is set." I believe although i was greate feature to be added, it is more consistent to use default value None rather than Lax. However this may have some impact, on upgrading from 1.1.22 to 1.1.23 as it will be by default a more loose rule

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

### DIFF
--- a/framework/web/CHttpCookie.php
+++ b/framework/web/CHttpCookie.php
@@ -80,9 +80,10 @@ class CHttpCookie extends CComponent
 	 * @var array Cookie attribute "SameSite".
 	 * @see https://www.owasp.org/index.php/SameSite
 	 * This property only works for PHP 7.3.0 or above.
+	 * Set this value to None because there might be breaking changes on different default value.
 	 * @since 1.1.22
 	 */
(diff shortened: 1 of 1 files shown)


samdark:
I don't think this default should be changed:

1. Lax is fine as default.
2. That's a backwards compatibility break.

nikolasr200:
I faced an issue upgrading to 7.3.x and yii 1.1.22 only to find out that it was cause by the Lax value. This is the reason i considered it would be more consistent to have None as a default value. 
Anyway, thank you for your time to review this 👍

samdark:
For Yii 2 it is lax by default as well in case PHP is more than 7.

Issue #4317, opened 2020-09-24
Added special option 'encode' to `$htmlOptions` argument in `CHtml::errorSummary` and `CHtml::error`


shidenko97:
Problem with HTML tags in validator's error text

I want to rollback changes in file **framework/web/helpers/CHtml.php** from [this commit](https://github.com/yiisoft/yii/commit/acb7a3b4881b693a1a19178694e4704038df9452) . It will make HTML tags in validator's error text available again.

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.22
| PHP version      | Any
| Operating system | Any

shidenko97:
It has at least 4 reasons to make it.

- https://github.com/yiisoft/yii/commit/acb7a3b4881b693a1a19178694e4704038df9452#r33834132
- https://github.com/yiisoft/yii/commit/acb7a3b4881b693a1a19178694e4704038df9452#r33767237

samdark:
I don't think it's a good idea:

1. It may introduce security issue.
2. It was done a long time ago so there are projects relying on the behavior.

We may introduce an option for htmlOptions to turn off HTML escaping though.

shidenko97:
> I don't think it's a good idea:
> 
> 1. It may introduce security issue.
> 2. It was done a long time ago so there are projects relying on the behavior.
> 
> We may introduce an option for htmlOptions to turn off HTML escaping though.

Yes, it looks like better idea than my. I've did it by your way.

PR #4318, opened 2020-09-24
Added special option 'encode' to `$htmlOptions` argument in `CHtml::errorSummary` and `CHtml::error`

shidenko97:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #4317

### DIFF
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -2220,7 +2220,7 @@ EOD;
 				foreach($errors as $error)
 				{
 					if($error!='')
-						$content.= '<li>'.self::encode($error)."</li>\n";
+						$content.= '<li>'.$error."</li>\n";
 					if($firstError)
 						break;
 				}
@@ -2251,7 +2251,7 @@ EOD;
 	public static function error($model,$attribute,$htmlOptions=array())
 	{
 		self::resolveName($model,$attribute); // turn [a][b]attr into attr
-		$error=self::encode($model->getError($attribute));
+		$error=$model->getError($attribute);
 		if($error!='')
 		{
 			if(!isset($htmlOptions['class']))
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -2197,10 +2197,10 @@ EOD;
 	 * make the error summary to show only the first error message of each attribute.
 	 * If this is not set or is false, all error messages will be displayed.
 	 * This option has been available since version 1.1.3.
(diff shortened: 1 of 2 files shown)


framework/web/helpers/CHtml.php
```diff
@@ -2220,7 +2224,11 @@ public static function errorSummary($model,$header=null,$footer=null,$htmlOption
 				foreach($errors as $error)
 				{
 					if($error!='')
-						$content.= '<li>'.$error."</li>\n";
+					{
+                        if (!isset($htmlOptions['ignoreHtmlEncode']) || !$htmlOptions['ignoreHtmlEncode'])
```
samdark, line 2228: Let's be consistent in naming: https://github.com/yiisoft/yii/blob/master/framework/web/helpers/CHtml.php#L153

shidenko97: Sorry, have not seen it. Few moments and I will fix it with tabs.

framework/web/helpers/CHtml.php
```diff
@@ -2197,6 +2197,10 @@ public static function activeRadioButtonList($model,$attribute,$data,$htmlOption
 	 * make the error summary to show only the first error message of each attribute.
 	 * If this is not set or is false, all error messages will be displayed.
 	 * This option has been available since version 1.1.3.
+	 * Another special option named 'encode' is recognized, which when set true, will
```
samdark, line 2200: It should be `false`:

```suggestion
	 * Another special option named 'encode' is recognized, which when set false, will
```

framework/web/helpers/CHtml.php
```diff
@@ -2197,6 +2197,10 @@ public static function activeRadioButtonList($model,$attribute,$data,$htmlOption
...
+	 * disable the CHtml::encode encoding of all error messages.
+	 * If this is not set or is false, all error messages will be encoded by CHtml::encode.
```
samdark, line 2202:
```suggestion
	 * If this is not set or is true, all error messages will be encoded by CHtml::encode.
```

samdark (changes requested):

framework/web/helpers/CHtml.php
shidenko97: Thanks, I've changed it and condition to correct

samdark (approved):

samdark:
👍

PR #4322, opened 2020-10-01
Getter methods available for attributes property

Antonio80negro:
It must be possible to override, in a child class, the value of an attribute property, to give developers more flexibility.
Now it's possible to define a new getter method in an ActiveRecord extended class, even for attributes property, in order to process original values saved in db and modify its values.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/framework/db/ar/CActiveRecord.php
+++ b/framework/db/ar/CActiveRecord.php
@@ -133,7 +133,10 @@ abstract class CActiveRecord extends CModel
 	 */
 	public function __get($name)
 	{
-		if(isset($this->_attributes[$name]))
+	 	$getter = 'get'.$name;
+            	if(method_exists($this, $getter))
+                	return $this->$getter();
+		elseif(isset($this->_attributes[$name]))
 			return $this->_attributes[$name];
 		elseif(isset($this->getMetaData()->columns[$name]))
 			return null;


viktorprogger:
This is a good point, I really tried to do this in my current project, but these changes break backwards compatibility very hard. All the current logic built on difference between `$record->attributeName` and `$record->getAttributeName()` in current projects will be broken (it was the reason I've reverted my changes after spending a lot of time on bugfixing).

Instead, I suggest you to inherit `CActiveRecord` in your projects with your own class with a such `__get` method:
```php
public function __get($name) 
{
    $getter = "get$name";
    if(method_exists($this, $getter)) {
        return $this->$getter();
    }

    return parent::__get($name);
}
```

Antonio80negro:
> 
> Instead, I suggest you to inherit `CActiveRecord` in your projects with your own class with a such `__get` method:
> 
> ```
> public function __get($name) 
> {
>     $getter = "get$name";
>     if(method_exists($this, $getter)) {
>         return $this->$getter();
>     }
> 
>     return parent::__get($name);
> }
> ```

It's exactly what i have done, but i wished to pull this solution in the framework as well! :)

viktorprogger:
> It's exactly what i have done, but i wished to pull this solution in the framework as well! :)

I don't think @yiisoft/core-developers will approve this PR as it breaks backwards compatibility :(

Antonio80negro:
Sorry!
I haven't understood which would be the brake in BC?
If the developer doesn't override a getter method for an attribute property, it won't change anything!
So BC must be kept, didn't it?

viktorprogger:
> If the developer doesn't override a getter method for an attribute property, it won't change anything!
You can't guarantee that there were no developers who created `getAttrubteName` method. My concrete project is a live example: it has lots and lots of functions in AR classes which doing completely different things (I really don't even wanna know why did somebody name them this way). This change will completely break all our system.

Antonio80negro:
Ok, that was the point! :) You were right!

samdark:
Won't accept it for the reasons @viktorprogger highlighted.

Issue #4302, opened 2020-07-02
Updated postgres 12 query for findColumns function returns wrong sequence name in the result


varunkamra:
Postgres 12 updated the findColumns function and the following query when run for a database table whose name and sequence name were updated after it was created returns old sequence name when this query gets executed.

`SELECT a.attname, LOWER(format_type(a.atttypid, a.atttypmod)) AS type,
	d.adsrc AS column_def_value, a.attnotnull, a.atthasdef,
	pg_catalog.col_description(a.attrelid, a.attnum) AS comment
FROM pg_attribute a LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
WHERE a.attnum > 0 AND NOT a.attisdropped
	AND a.attrelid = (SELECT oid FROM pg_catalog.pg_class WHERE relname=:table
		AND relnamespace = (SELECT oid FROM pg_catalog.pg_namespace WHERE nspname = :schema))
ORDER BY a.attnum`

Old query that gets generated by findColumns function before postgres12 upgrade code was added gives the right sequence name for the same table

`SELECT a.attname, LOWER(format_type(a.atttypid, a.atttypmod)) AS type, CAST(pg_get_expr(d.adbin, d.adrelid) AS varchar) AS adsrc, a.attnotnull, a.atthasdef,
	pg_catalog.col_description(a.attrelid, a.attnum) AS comment
FROM pg_attribute a LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
WHERE a.attnum > 0 AND NOT a.attisdropped
	AND a.attrelid = (SELECT oid FROM pg_catalog.pg_class WHERE relname=:table
		AND relnamespace = (SELECT oid FROM pg_catalog.pg_namespace WHERE nspname = :schema))
ORDER BY a.attnum`

**PHP version being used** 7.2.31
**Postgres** 11.8
**OS**: CentOS Linux release 7.7.1908 (Core), Ubuntu 20.04 LTS

https://github.com/yiisoft/yii/blob/09672fe0754eb939bafaf28a13b9dcad4bfaca0e/framework/db/schema/pgsql/CPgsqlSchema.php#L172

samdark:
Have you tried with dev-master version? https://github.com/yiisoft/yii/blob/master/CHANGELOG#L7

varunkamra:
[...]

PR #4323, opened 2020-10-02
Add postgresql 12 support

marcovtwout:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | **requires confirmation**
| Fixed issues  | https://github.com/yiisoft/yii/issues/4294, https://github.com/yiisoft/yii/issues/4302

This PR replaces the implementation in https://github.com/yiisoft/yii/issues/4297 with an improved implementation according to the notes in https://www.postgresql.org/docs/12/release-12.html:

> - Remove obsolete pg_constraint.consrc column (Peter Eisentraut)
> This column has been deprecated for a long time, because it did not update in response to other catalog changes (such as column renamings). The recommended way to get a text version of a check constraint's expression from pg_constraint is pg_get_expr(conbin, conrelid). pg_get_constraintdef() is also a useful alternative.

> - Remove obsolete pg_attrdef.adsrc column (Peter Eisentraut)
> This column has been deprecated for a long time, because it did not update in response to other catalog changes (such as column renamings). The recommended way to get a text version of a default-value expression from pg_attrdef is pg_get_expr(adbin, adrelid).

The new implementation should be fully backwards compatible on older versions of PostgreSQL down to the oldest supported version 9.5 according to https://www.postgresql.org/support/versioning/

### DIFF
--- a/framework/db/schema/pgsql/CPgsqlSchema.php
+++ b/framework/db/schema/pgsql/CPgsqlSchema.php
@@ -169,7 +169,7 @@ class CPgsqlSchema extends CDbSchema
 SELECT
 	a.attname,
 	LOWER(format_type(a.atttypid, a.atttypmod)) AS type,
-	pg_get_expr(adbin, adrelid) AS default_value_text,
+	pg_get_expr(adbin, adrelid) AS adsrc,
 	a.attnotnull,
 	a.atthasdef,
 	pg_catalog.col_description(a.attrelid, a.attnum) AS comment
@@ -191,7 +191,7 @@ EOD;
 			$c=$this->createColumn($column);
 			$table->columns[$c->name]=$c;
 
-			if(stripos($column['default_value_text'],'nextval')===0 && preg_match('/nextval\([^\']*\'([^\']+)\'[^\)]*\)/i',$column['default_value_text'],$matches))
+			if(stripos($column['adsrc'],'nextval')===0 && preg_match('/nextval\([^\']*\'([^\']+)\'[^\)]*\)/i',$column['adsrc'],$matches))
 			{
 				if(strpos($matches[1],'.')!==false || $table->schemaName===self::DEFAULT_SCHEMA)
 					$this->_sequences[$table->rawName.'.'.$c->name]=$matches[1];
@@ -218,7 +218,7 @@ EOD;
 		$c->isForeignKey=false;
 		$c->comment=$column['comment']===null ? '' : $column['comment'];
 
-		$c->init($column['type'],$column['atthasdef'] ? $column['default_value_text'] : null);
+		$c->init($column['type'],$column['atthasdef'] ? $column['adsrc'] : null);
 
 		return $c;
 	}
--- a/framework/db/schema/pgsql/CPgsqlSchema.php
+++ b/framework/db/schema/pgsql/CPgsqlSchema.php
@@ -166,7 +166,12 @@ class CPgsqlSchema extends CDbSchema
 	protected function findColumns($table)
 	{
 		$sql=<<<EOD
-SELECT a.attname, LOWER(format_type(a.atttypid, a.atttypmod)) AS type, d.adsrc, a.attnotnull, a.atthasdef,
+SELECT
+	a.attname,
+	LOWER(format_type(a.atttypid, a.atttypmod)) AS type,
+	d.adsrc,
+	a.attnotnull,
+	a.atthasdef,
 	pg_catalog.col_description(a.attrelid, a.attnum) AS comment
 FROM pg_attribute a LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
 WHERE a.attnum > 0 AND NOT a.attisdropped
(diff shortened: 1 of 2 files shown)


framework/db/schema/pgsql/CPgsqlSchema.php
```diff
@@ -166,7 +166,12 @@ protected function resolveTableNames($table,$name)
 	protected function findColumns($table)
 	{
 		$sql=<<<EOD
-SELECT a.attname, LOWER(format_type(a.atttypid, a.atttypmod)) AS type, d.adsrc, a.attnotnull, a.atthasdef,
+SELECT
+	a.attname,
+	LOWER(format_type(a.atttypid, a.atttypmod)) AS type,
+	pg_get_expr(adbin, adrelid) AS default_value_text,
```
cebe, line 172: keeping the name `adsrc` here would make this fully BC save, in case someone extended that class.

```suggestion
	pg_get_expr(adbin, adrelid) AS adsrc,
```

marcovtwout, line 172: @cebe Do you still think the old name "adsrc" should be kept as an alias? It is only used inside of this function and in createColumn().

marcovtwout: Added to this PR

marcovtwout:
Side note, this implementation now basically matches an earlier PR that also looks correct: https://github.com/yiisoft/yii/pull/4296/files

framework/db/schema/pgsql/CPgsqlSchema.php
cebe: yes, createColumn() could be extended.

marcovtwout:
Some extra notes about backwards compatibility:

PostgreSQL version 9.5 is the current oldest supported version, and 7.3 is the version still mensioned in Yii1 docs https://www.yiiframework.com/doc/guide/1.1/en/database.ar)

The adsrc field was already mensioned as "historical" in the docs of version 8: https://www.postgresql.org/docs/8.0/catalog-pg-attrdef.html. And the consrc field was already discouraged on version 7.3: https://www.postgresql.org/docs/7.4/catalog-pg-constraint.html. So in my opinion the new implementation should fully replace the old implementation.

marcovtwout:
https://github.com/yiisoft/yii/pull/4297 must be reverted before this can be properly merged.

samdark:
@marcovtwout maybe add a commit reverting it?

marcovtwout:
@samdark More clean seems to be revert button on the previous PR, which creates https://github.com/yiisoft/yii/pull/4324. When that one is merged, this one can be merged.

samdark:
> So in my opinion the new implementation should fully replace the old implementation.

That would be a backwards compatibility break and we don't really want that. Is it possible to support both versions?

samdark:
I've checked how it was done in https://github.com/yiisoft/yii2/blob/master/framework/db/pgsql/Schema.php and it seems old versions support new syntax well.

samdark:
@marcovtwout in case you've tested it and it works, feel free to merge it.

cebe (approved):

samdark:
@marcovtwout let's resolve conflicts and merge it.

marcovtwout:
Have asked a few contributors to test and confirm. I'll also update the changelog before merging.

batyrmastyr:
@marcovtwout looks like you missed [issue](https://github.com/yiisoft/yii2/commit/09eacd8dc1120113032976099fd07290a4d37cd3#diff-03b64c04d37c13a3118b4b8119f5ae8dR596) with `GENERATED BY IDENTITY` columns as their default value is NULL therefore [this](https://github.com/yiisoft/yii/pull/4323/files#diff-2c605058d2eadac7c84bab33adbe5a7fR194-R201) code won't execute.

marcovtwout:
@batyrmastyr Thanks for your feedback.

The point about `GENERATED BY IDENTITY` seems to be a new feature introduced in PostgreSQL 10. This PR is only about adding compatibility support for v12 so you can upgrade without breaking your application. Do you see anything else missing that is strictly required for compatibility?

@samdark We could consider porting more functionality from Yii 2 / https://github.com/yiisoft/yii2/issues/17597, but I think that would be better suited in a separate PR.

batyrmastyr:
@marcovtwout yes, `GENERATED BY IDENTITY` is a new feature of PG 12, but it breaks `$column->autoIncrement` and `$schema->resetSequence`. The catch is that under the hood generated identity column has an ordinary `sequence`, just like a `serial` columns.
Still, I'm not a contributor and I don't think it's a critical bug.

marcovtwout:
@batyrmastyr It was already added in v10 if I understand correctly. so that must have been broken for a while. I guess we could log that as a known issue with the Yii1 implementation and continue.

marcovtwout:
Got confirmation in https://github.com/yiisoft/yii/issues/4294#issuecomment-705049110 and  https://github.com/yiisoft/yii/issues/4302#issuecomment-712159510

I'll update the changelog and merge.

PR #4340, opened 2020-12-28
fix #4339

### LINKED ISSUE #4339: PHP 8 - "There is no active transaction" when transaction is autocommitted
This is related to https://github.com/yiisoft/yii2/issues/18406 

As of PHP 8, PDO::commit() and rollBack() will throw a There is no active transaction PDOException if the transaction had been autocommitted, as MySQL will do whenever a schema change is made. Previously they would fail silently.

1. Connect to a MySQL database
2. Call Yii::app()->db->beginTransaction()
3. Make a schema change, such as add a new column to a table (the transaction will be autocommitted in the process)
4. Call Yii::app()->db->commit()

PHP would move along to the next line of code, regardless of whether the app is running on PHP 5, 7, or 8.

### What do you get instead?
A PDOException is thrown with the message There is no active transaction when running PHP 8. 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.24-dev
| PHP version      | 8.0.0
| Operating system | Centos 7

twisted1919:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4339

### DIFF
--- a/framework/db/CDbTransaction.php
+++ b/framework/db/CDbTransaction.php
@@ -61,7 +61,8 @@ class CDbTransaction extends CComponent
 		if($this->_active && $this->_connection->getActive())
 		{
 			Yii::trace('Committing transaction','system.db.CDbTransaction');
-			$this->_connection->getPdoInstance()->commit();
+			if($this->_connection->getPdoInstance()->inTransaction())
+				$this->_connection->getPdoInstance()->commit();
 			$this->_active=false;
 		}
 		else
@@ -77,7 +78,8 @@ class CDbTransaction extends CComponent
 		if($this->_active && $this->_connection->getActive())
 		{
 			Yii::trace('Rolling back transaction','system.db.CDbTransaction');
-			$this->_connection->getPdoInstance()->rollBack();
+			if($this->_connection->getPdoInstance()->inTransaction())
+				$this->_connection->getPdoInstance()->rollBack();
 			$this->_active=false;
 		}
 		else
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,7 @@
 Version 1.1.24 under development
 --------------------------------
 
-- no changes yet.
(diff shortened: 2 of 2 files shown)


samdark (approved):

samdark:
Thanks!

framework/db/CDbTransaction.php
```diff
@@ -61,7 +61,8 @@ public function commit()
 		if($this->_active && $this->_connection->getActive())
 		{
 			Yii::trace('Committing transaction','system.db.CDbTransaction');
-			$this->_connection->getPdoInstance()->commit();
+			if($this->_connection->getPdoInstance()->inTransaction())
```
marcovtwout, line 64: @samdark Shouldn't the trace message be part of the if-block?

samdark: Yes, I think so.

twisted1919:
I can adjust this thing later this month if you wish and do another pull request, but for now, i simply don't have enough spare time to do it. 

If it's not acceptable, @marcovtwout can you please take care of it?

marcovtwout:
@twisted1919 I've reviewed this change further and see that it is completely in line with the change in Yii 2 (https://github.com/yiisoft/yii2/pull/18407/files). The Yii 2 implementation also logs the debug/trace even if the transaction was already autocommitted, so I suggest we leave it as is.

twisted1919:
Awesome :)

samdark:
Fixed the tag. Thanks. @marcovtwout any ETA on release?

marcovtwout:
@forevermatt I'm planning a regular release schedule, the next release should be no later then 2020-06-01. You can use master in the meantime.

forevermatt:
That makes sense. Thanks!

PR #4344, opened 2021-02-08
Upgrade jQuery and jQuery UI

marcovtwout:
| Q             | A
| ------------- | ---
| Is bugfix?    | n/a
| New feature?  | n/a
| Breaks BC?    | ✔️/❌?
| Tests pass?   | ✔️
| Fixed issues  | 

Updated the bundled jQuery and jQuery UI to the latest 1.x versions. They are already a few years old and pretty stale, but I think it makes sense to update them.

jQuery UI v12 contains one important BC question to decide on: a new default theme. I think it is best to use the new default theme as default and provide a way to revert to the old theme, see UPGRADE.md.


samdark:
Are these the only BC breaks?

marcovtwout:
@samdark the jQuery update contains no breaking changes.

The jQuery UI update might contain a few more. I did a quick test rendering all the CJuiWidgets in the base application template, and the changed theme was the only one that stood out. I also scanned the full changelog, and while they marked a lot of stuff deprecated, not much was removed and the removed functions are not found in yii1 core code. The full changelog is here: https://jqueryui.com/upgrade-guide/1.12/#general-changes and I included it in UPGRADE.md.

samdark:
> and the removed functions are not found in yii1 core code

That doesn't mean these aren't used in project code.

marcovtwout:
> That doesn't mean these aren't used in project code.
Of course, that's why I linked to the full changelog in the UPGRADE file, just like we did when releasing 1.1.15 (https://github.com/yiisoft/yii/blob/master/UPGRADE#L98)

Do you think more should be done?

samdark:
Well, that is, technically, a backwards compatibility break so either we need to emphasize it (as it was done in UPGRADE) or just give up the idea.

twisted1919:
Whomever needs to upgrade core jquery/jquery ui can do so outside the framework very easily. I don't think doing a BC for this actually worth. It just adds extra work for nothing in return, which cannot be done outside the framework.

rob006:
AFAIK both jQuery and jQuery UI has known vulnerabilities in 1.11.x line, so "nothing in return" is not appropriate here.

samdark:
Tests are alright now.

marcovtwout:
@rob006 Since jQuery 1.12 does not contain breaking changes, there should be no objections there. And although perhaps minor, the known XSS vulnerability in jQuery UI is reason enough for me to push that upgrade as well.

Main questions that remains: should we also update the base theme to the new default? (with the added instruction in https://github.com/yiisoft/yii/pull/4344/commits/37f808df46de104ccfe3787141480d9d385ed35f). For me this makes the most sense as it is the external library that dictates the defaults, and not Yii.

rob006:
> Main questions that remains: should we also update the base theme to the new default? (with the added instruction in [37f808d](https://github.com/yiisoft/yii/commit/37f808df46de104ccfe3787141480d9d385ed35f)). For me this makes the most sense as it is the external library that dictates the defaults, and not Yii.

If library is bundled in framework, it is framework responsibility in terms of BC and security. If we can keep BC in this case, we should keep it and use previous theme by default.

samdark:
I'd keep the old theme.

marcovtwout:
That is fine by me as well, I'll update the code and switch the default and instructions for BC.

samdark (approved):

PR #4347, opened 2021-02-18
Fix GitHub actions

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ?
| Fixed issues  | -


samdark:
Seems to be connected with recent PHP action release: https://github.com/shivammathur/setup-php/issues/412

Issue #4349, opened 2021-03-04
Remove attribute tag from <script> tag


mohamedmalki:
### What steps will reproduce the problem?

Hi, There is any way to remove or not use Attribute 'type="text/javascript"' when we use create a new js file usinf one of this function ? :
Yii::app()->clientScript->registerScript('myjavascript','(function() { // code here })();',CClientScript::POS_END);
or
Yii::app()->clientScript->registerScriptFile( $file,CClientScript::POS_END );

### What is the expected result?

It gives  '<script type="text/javascript"> // code here</script>', and W3C validator says that is a warning, you have to remove the type attribute.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.19
| PHP version      | 7.1
| Operating system |

samdark:
Yes, should be adjusted.

marcovtwout:
Technically it's not wrong to keep this obsolete attribute, but it's a small change to keep the w3c validator happy. I don't see any objections against removing it, unless there would be a known component in any way dependent on it. PR coming up.

marcovtwout:
It seems we can't simply remove it always for backwards compatibility with XHTML. Lets continue discussion in the PR.

marcovtwout:
@mohamedmalki Take a look at the solution in https://github.com/yiisoft/yii/pull/4351

marcovtwout:
Fixed with https://github.com/yiisoft/yii/pull/4351/

You have to enable the new option to use this functionality.

PR #4350, opened 2021-03-08
Remove script type attribute

marcovtwout:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️/❌?
| Tests pass?   | ✔️
| Fixed issues  | #4349

### DIFF
--- a/framework/gii/views/layouts/main.php
+++ b/framework/gii/views/layouts/main.php
@@ -25,7 +25,7 @@ $cs->registerCssFile($baseUrl.'/js/fancybox/jquery.fancybox-1.3.1.css');
 
 	<title><?php echo CHtml::encode($this->pageTitle); ?></title>
 
-	<script type="text/javascript" src="<?php echo $this->module->assetsUrl; ?>/js/main.js"></script>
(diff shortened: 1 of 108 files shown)


rob006:
Shouldn't this be configurable? In the same way as `Html::$closeSingleTags` and similar flags?

marcovtwout:
@rob006 Did not know about those flags, but that is an option. It would be a bit harder to implemented since Yii also uses hardcoded view files that include this attribute.

I just noticed the previous XHTML standard still requires the type attribute, so I guess we have to at least keep it for backwards compatibility.

rob006:
> It would be a bit harder to implemented since Yii also uses hardcoded view files that include this attribute.

If they have doctype defined as XHTML, it should not be changed. Other partial views seems to be debug-related, so I wouldn't bother to change these. In general I would not change anything except `CHtml`, and maybe `CWebLogRoute` to use `CHtml::script()` to inherit settings from helper.

There is also `/*<![CDATA[*/` which seems to be redundant in HTML5.

marcovtwout:
Closing this PR in favor of https://github.com/yiisoft/yii/pull/4351

PR #4351, opened 2021-03-11
Options for html5 script and style

marcovtwout:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4349

Added options based on further discussion in https://github.com/yiisoft/yii/pull/4350

@rob006 Please review if you want :) I've opted to only change CHtml and keep the default/logging views (including CWebLogRoute) the same.

### DIFF
--- a/tests/framework/web/helpers/CHtmlTest.php
+++ b/tests/framework/web/helpers/CHtmlTest.php
@@ -328,6 +328,20 @@ class CHtmlTest extends CTestCase
 	{
 		$this->assertEquals($assertion, CHtml::script($text));
 	}
+	
+	/**
+	 * @dataProvider providerScript
+	 * @backupStaticAttributes enabled
+	 *
+	 * @param string $text
+	 * @param string $assertion
+	 */
+	public function testScriptHtml5($text, $assertion)
+	{
+		CHtml::$setScriptType=false;
+		$assertion=str_replace(' type="text/javascript"', '', $assertion);
+		$this->assertEquals($assertion, CHtml::script($text));
+	}
 
 	public static function providerScriptWithHtmlOptions()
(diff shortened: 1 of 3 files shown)


tests/framework/web/helpers/CHtmlTest.php
```diff
@@ -328,6 +352,42 @@ public function testScript($text, $assertion)
...
+	/**
+	 * @dataProvider providerScript
+	 * @backupStaticAttributes enabled
+	 *
+	 * @param string $text
+	 * @param string $assertion
+	 */
+	public function testScriptHtml5($text, $assertion)
+	{
+		CHtml::$setScriptType=false;
```
rob006, line 387: Shouldn't we reset this flag after end of the test?

rob006 (approved):

samdark (approved):

tests/framework/web/helpers/CHtmlTest.php
marcovtwout: That is what the `@backupStaticAttributes enabled` annotation should be taking care of. (https://phpunit.readthedocs.io/en/9.5/annotations.html#backupstaticattributes). But note that this is a very broad approach - it backups and resets any static class properties.

Is it perhaps better to explicitly reset the required static attributes in_setUp() ?

marcovtwout: PHPUnit always has multiple ways to achieve the same thing. :) I try to adhere to the existing testing style as much as possible and not introduce new concepts, but this is used in a few other places as well.

PR #4354, opened 2021-04-07
Added log file permissions

jdayamx:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@ Version 1.1.24 under development
 - Bug #4355: Fix errorhandler missing backtrace entries (georaldc, marcovtwout)
 - Enh #4349: Added CHtml option to omit type attribute from <script> tag (mohamedmalki, marcovtwout)
 - Enh #4351: Added CHtml option to omit CDATA wrapper from <script> and <style> contents (marcovtwout)
+- Enh #4354: Allow to set log file permissions for CFileLogRoute (jdayamx)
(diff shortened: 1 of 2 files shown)


marcovtwout:
@jdayamx Thanks for your PR. Yii 1 is in maintainance mode, but if it can be added in a backwards compatible manner we can consider it for merging.

This PR is currently backward incompatible. Default behavior relies on system settings (such as umask) to set permissions. This PR overwrites that with 0644.

Take a look at the implementation done by Yii 2 for `$fileMode`: https://github.com/yiisoft/yii2/blob/master/framework/log/FileTarget.php - if you apply behavior and documentation exactly like that I'll be happy to merge it.

jdayamx:
I got such a situation that the console script and the application were launched under different users, but there was  hase one group, and when the file was created by one user, the second could not write anything to the log. For the file system and working with files, developers should be able to manage file permissions without crutches with inheritance.

marcovtwout:
The PR in its current form still sets default file permissions and this breaks backwards compatibility. Take a closer look at the Yii 2 implementation mentioned above if you would like to get this merged.

marcovtwout (approved):

marcovtwout:
Update documentation a bit and added changelog entry. Merged, thanks for your contribution!

Issue #4226, opened 2018-09-17
reset() expects parameter 1 to be array, integer given in GII diff


ahmadasjad:
Note that only PHP 7 compatibility issues are accepted. For security issues contact maintainers privately.

### What steps will reproduce the problem?

[model_and_controller_to_check_yii1.1.20_gii_error.zip](https://github.com/yiisoft/yii/files/2388211/model_controller_to_check_yii1.1.20_gii_error.zip)
There are two files one is a controller and another is a model. Once you'll try to regenerate the crud from the Model, click **preview** button then click on **diff** text of the controller going to be generated. Now you'll see the error `reset() expects parameter 1 to be array, integer given`

### What is the expected result?
Difference between old and new code when going to generate crud from Active Record model using GII

### What do you get instead?
reset() expects parameter 1 to be array, integer given

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.20.6ed384
| PHP version      | 7.1.20-1
| Operating system | ubuntu16.04.1

ahmadasjad:
### Here is the stacktrace:
```
2018/09/17 15:40:10 [error] [php] reset() expects parameter 1 to be array, integer given (/var/www/html/yii-1.1.20.6ed384/framework/gii/components/Pear/Text/Diff/Engine/native.php:196)
Stack trace:
#0 /var/www/html/yii-1.1.20.6ed384/framework/gii/components/Pear/Text/Diff/Engine/native.php(88): Text_Diff_Engine_native->_compareseq()
#1 /var/www/html/yii-1.1.20.6ed384/framework/gii/components/Pear/Text/Diff.php(56): Text_Diff_Engine_native->diff()
#2 /var/www/html/yii-1.1.20.6ed384/framework/gii/components/Pear/Text/Diff/Renderer/inline.php(136): Text_Diff->__construct()
#3 /var/www/html/yii-1.1.20.6ed384/framework/gii/components/Pear/Text/Diff/Renderer.php(165): Text_Diff_Renderer_inline->_changed()
#4 /var/www/html/yii-1.1.20.6ed384/framework/gii/components/Pear/Text/Diff/Renderer.php(140): Text_Diff_Renderer_inline->_block()
#5 /var/www/html/yii-1.1.20.6ed384/framework/gii/components/TextDiff.php(20): Text_Diff_Renderer_inline->render()
[...]

PR #4373, opened 2021-09-15
Fix "reset() expects parameter 1 to be array" raised by Gii diff, issue #4226

LeoZandvliet:
Fix "reset() expects parameter 1 to be array" raised by Gii diff, issue #4226

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️/❌
| Fixed issues  | #4226

### DIFF
--- a/framework/gii/components/Pear/Text/Diff/Engine/native.php
+++ b/framework/gii/components/Pear/Text/Diff/Engine/native.php
@@ -191,31 +191,17 @@ class Text_Diff_Engine_native {
                     continue;
                 }
                 $matches = $ymatches[$line];
-                reset($matches);
-                foreach($matches as $match) {
-                    reset($match);
-                    $y = next($match);
+                foreach ($matches as $y) {
                     if (empty($this->in_seq[$y])) {
                         $k = $this->_lcsPos($y);
                         assert($k > 0);
                         $ymids[$k] = $ymids[$k - 1];
                         break;
-                    }
-                }
-                foreach($matches as $match) {
-                    reset($match);
-                    $y = next($match);
-                    if ($y > $this->seq[$k - 1]) {
+                    } elseif ($y > $this->seq[$k - 1]) {
                         assert($y <= $this->seq[$k]);
-                        /* Optimization: this is a common case: next match is
-                         * just replacing previous match. */
                         $this->in_seq[$this->seq[$k]] = false;
                         $this->seq[$k] = $y;
                         $this->in_seq[$y] = 1;
-                    } elseif (empty($this->in_seq[$y])) {
-                        $k = $this->_lcsPos($y);
-                        assert($k > 0);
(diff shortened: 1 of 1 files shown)


LeoZandvliet:
@marcovtwout I added a test file to #4226 which you can use to verify the bug and my pull request: https://github.com/yiisoft/yii/issues/4226#issuecomment-920110124

marcovtwout:
Testing this patch in a local project with assertions enabled throws:

```
assert($y <= $this->seq[$k]) ((..)yiisoft\yii\framework\gii\components\Pear\Text\Diff\Engine\native.php:201)
```

It's a pretty complicated bit of code. After examining the horde patch (https://github.com/horde/Text_Diff/commit/65f92a945f3745be2953d2bcc24735e860cf70a7) some more, it seems they refactored `each()` incorrectly by combining both loops into one. Notice how in the original code the first loop sets `$k` and breaks when found. Then the the second loop continues using that value of `$k`. When both loops are combined `$k` is used while still undefined and then the assertion fails.

When I refactor the original each() calls to foreach() while keeping the structure intact, it seems to work as expected (tested on PHP 7.3 and 8.0). Alternative PR coming up.

marcovtwout:
@LeoZandvliet Could you test https://github.com/yiisoft/yii/pull/4377?

LeoZandvliet:
Hey @marcovtwout , what input do you use to trigger the assertions? In both of the PR I'm unable to get them (using the tests I've written and attached #4226). 
Also, as I don't understand the code you changed on top of the horde patch, in what cases and what kind of difference would you see in the 'diff output' ? 
In either case, this PR works :)

marcovtwout:
@LeoZandvliet I extracted a minimal test case from Gii model generation in a project I was testing in:

```php
Yii::import('gii.components.TextDiff');
		
$a = <<<'PHP'
class User extends CActiveRecord
{
    public static function model($className=__CLASS__)
    {
        return parent::model($className);
    }
	
    public function relations()
    {
        return [];
    }
}
PHP;

$b = <<<'PHP'
class User extends CActiveRecord
{
	/**
	 * @return array relational rules.
	 */
	public function relations()
	{
		return [];
	}

	public static function model($className=__CLASS__)
	{
		return parent::model($className);
	}
}
PHP;

var_dump(TextDiff::compare($a, $b));
```

Results in:
```
AssertionError
assert($y <= $this->seq[$k])

(..)\yiisoft\yii\framework\gii\components\Pear\Text\Diff\Engine\native.php(204)
```

It turns out my solution in #4377 also doesn't work. I've misunderstood how the loops work: the original using `each()` does not reset the array pointer in between, so that refactor to foreach() is wrong. I cannot continue debugging at the moment but will revisit this later. Or perhaps you can try the above test case as well and see if you spot the correct way to refactor this?

PR #4374, opened 2021-09-22
Fixed issue: createUpdateCommand does not accept just a table name when using MSSQL

c-schmitz:
If a tablename as string is given the following error occurs: 
Error: Call to a member function getColumn() on string 

Normal CDBCommand (Mysql) does this just fine.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  |❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,8 +5,7 @@ Version 1.1.25 under development
 --------------------------------
 
 - Bug #4369: PHP 8.0 compatibility: Fix warning "Only the first byte will be assigned to the string offset" when generating code with Gii (marcovtwout)
-- Bug: createUpdateCommand does not accept just a table name when using MSSQL (c-schmitz)
-
+- Bug #4374: Fix for createUpdateCommand which did not accept just a table name when using MSSQL (c-schmitz)
 
 Version 1.1.24 June 7, 2021
 --------------------------------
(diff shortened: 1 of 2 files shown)


marcovtwout:
That function `ensureTable()` only assets if the tablename is valid, and throws an exception if it is not. That does not seem to fix the issue you are describing?

c-schmitz:
Look at ensureTable again. It modifies the $table variable, and replaces it with an object if it is a string.

marcovtwout:
@c-schmitz You are correct :) Are there any other functions for MsSql where this should also be applied?

If you would update the CHANGELOG as well in your PR I'd be happy to merge it.

c-schmitz:
We use MSSQL extensively for years in Yii1 and did not encounter any other issues.
I wil update the CHANGELOG.

CHANGELOG
```diff
@@ -5,6 +5,8 @@ Version 1.1.25 under development
 --------------------------------
 
 - Bug #4369: PHP 8.0 compatibility: Fix warning "Only the first byte will be assigned to the string offset" when generating code with Gii (marcovtwout)
+- Bug: createUpdateCommand does not accept just a table name when using MSSQL (c-schmitz)
```
marcovtwout, line 8: Please include the PR number #4374 and start the sentence with a capital
(or allow maintainer edits and I'll fix that :) )

marcovtwout (approved):
Have not tested but static code looks good.

CHANGELOG
c-schmitz: Fixed it :-)

marcovtwout (approved):

marcovtwout:
Thanks! Merging

PR #4375, opened 2021-09-24
Increase validation string length size for decimal values

joshbmarshall:
When validating decimal fields e.g. currency, the string length is one character short of what can fit in the database. So need to add one to take the '.' into account

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |


rob006:
`extractLimit()` is overwritten in DBMS-specific classes (like `CPgsqlColumnSchema`), this should be also adjusted there.

marcovtwout:
I don't think this is the correct place to fix this, see the answer in https://github.com/yiisoft/yii/issues/4376

PR #4378, opened 2021-10-05
Use `callable` as type name instead of `callback` in phpdoc

rob006:
`callback` is not recognized by some SCA tools (like psalm) and treated as class named "callback". `callable` is internal PHP type and it should not have such problems (at least it fixed some psalm complains).

https://www.php.net/manual/en/language.types.callable.php


samdark (approved):

samdark:
Thanks.

PR #4457, opened 2022-08-04
Suport stripos null parameter deprecation that fixes #4456

### LINKED ISSUE #4456: PHP 8.1 CPgsqlSchema error stripos(): Passing null to parameter #1 ($haystack) of type string is deprecated
with Postgres Database configured
$model->find();

### What do you get instead?
an error: stripos(): Passing null to parameter #1 ($haystack) of type string is deprecated
CPgsqlSchema.php:194

### Additional info
adsrc is null when there's not a column default value, and stripos can't receive null

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.25
| PHP version      | 8.1
| Operating system | macOS

JonathanArgentao:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4456


marcovtwout (approved):
Thank you. I will squash the changelog messages before next release.

PR #4461, opened 2022-08-08
Param fix

olleharstedt:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | Param issue in function docblock


olleharstedt:
`itemAt` accepts `mixed`, so this change makes sense.

marcovtwout:
The class implements https://www.php.net/manual/en/class.arrayaccess.php which by now defines the param as `$mixed`. Is this the excepion in the source code or are there more similar changes to be made?

olleharstedt:
> The class implements https://www.php.net/manual/en/class.arrayaccess.php which by now defines the param as `$mixed`. Is this the excepion in the source code or are there more similar changes to be made?

No idea, honestly. Want me to grep around for array access interface?

olleharstedt:
Pushed some more fixes. Skipped CCache, since it's only accepting `string`.

olleharstedt:
@marcovtwout This can be merged :)

samdark (approved):

PR #4469, opened 2022-08-16
Add unix socket connection ability to CRedisCache

Catty2014:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | None

### DIFF
--- a/framework/caching/CRedisCache.php
+++ b/framework/caching/CRedisCache.php
@@ -85,15 +85,17 @@ class CRedisCache extends CCache
 	 */
 	protected function connect()
 	{
-        if ($this->socket)
-        {
+        $address = $this->hostname;
+        if (isset($this->port)){
+            $address .= ':' . $this->port;
+        }
         $this->_socket=@stream_socket_client(
-                $this->hostname.':'.$this->port,
-                $errorNumber,
-                $errorDescription,
-                $this->timeout ? $this->timeout : ini_get("default_socket_timeout"),
-                $this->options
-		);
+            $address,
+            $errorNumber,
+            $errorDescription,
+            $this->timeout ? $this->timeout : ini_get("default_socket_timeout"),
+            $this->options
+        );
 		if ($this->_socket)
 		{
 			if($this->ssl)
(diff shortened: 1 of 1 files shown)


framework/caching/CRedisCache.php
```diff
@@ -52,6 +53,10 @@ class CRedisCache extends CCache
 	 * @var int the port to use for connecting to the redis server. Default port is 6379.
 	 */
 	public $port=6379;
+	/**
+	 * @var string the unix socket to use for connecting to the redis server. If not set, hostname will be used.
+	 */
+	public $socket;
```
marcovtwout, line 59: This var name can be confusing since we also have $_socket.

framework/caching/CRedisCache.php
```diff
@@ -85,13 +90,26 @@ class CRedisCache extends CCache
...
+        else
+        {
+            $this->_socket=@stream_socket_client(
+                $this->hostname.':'.$this->port,
+                $errorNumber,
+                $errorDescription,
+                $this->timeout ? $this->timeout : ini_get("default_socket_timeout"),
+                $this->options
 		);
+        }
```
marcovtwout, line 112: A lot of duplicate code is introduced here, the first parameter is the only one swapped.

Instead of introducing a new parameter we could also do:

```php
$address = $this->hostname;
if (isset($this->port)) {
    $address .= ':' . $this->port;
}

$this->_socket=@stream_socket_client(
    $address,
    $errorNumber,
    (...)
```

And then you specify it like:
```php
 'cache'=>array(
    'class'=>'CRedisCache',
    'hostname'=>'unix://run/redis/redis.sock',
    'port'=>null
),
```

framework/caching/CRedisCache.php
```diff
@@ -85,13 +90,26 @@ class CRedisCache extends CCache
 	 */
 	protected function connect()
 	{
-		$this->_socket=@stream_socket_client(
-			$this->hostname.':'.$this->port,
-			$errorNumber,
-			$errorDescription,
-			$this->timeout ? $this->timeout : ini_get("default_socket_timeout"),
-			$this->options
+        if ($this->socket)
```
marcovtwout, line 93: Indentation and formatting does not adhere Yii coding standard (tabs)

marcovtwout (changes requested):

framework/caching/CRedisCache.php
rob006: Passing socket path as `hostname` is extremely confusing. I suggest to solve this in the same way as in Yii2: 

https://github.com/yiisoft/yii2-redis/blob/f96b872fa646308ab2f11721a7d35c713a234621/src/Connection.php#L283

marcovtwout (changes requested):
Sorry, this is currently not good enough as common framework code. Please check @rob006 comment, a better implementation is provided in Yii 2 and it could be made exactly like this: https://github.com/yiisoft/yii/pull/4469#discussion_r946494847
Also it needs to be properly documented before I can approve this,

PR #4478, opened 2022-08-25
PHP 8.1 support for active checkbox list

csears123:
In PHP 8.1 I am receiving the error below when using CActiveForm checkbox field, when the CFormModel model attribute has a default 'null' value as defined. 

"strcmp(): Passing null to parameter #2 ($string2) of type string is deprecated"

The 'null' value is passed to the checkBoxList() method as the second argument variable $select.  A null value is not allowed in PHP 8.1 in strcmp().  An additional condition check for !empty() on the $select variable will prevent this error, as well as support the remaining variable types for a string or array.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌

### DIFF
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -1168,7 +1168,7 @@ class CHtml
 
 		foreach($data as $value=>$labelTitle)
 		{
-			$checked=!is_array($select) && !strcmp($value, (string)$select) || is_array($select) && in_array($value,$select);
+			$checked=!is_array($select) && !strcmp($value,(string)$select) || is_array($select) && in_array($value,$select);
 			$checkAll=$checkAll && $checked;
 			$htmlOptions['value']=$value;
 			$htmlOptions['id']=$baseID.'_'.$id++;
@@ -1280,7 +1280,7 @@ EOD;
 		$id=0;
 		foreach($data as $value=>$labelTitle)
 		{
-			$checked=!strcmp($value, (string)$select);
+			$checked=!strcmp($value,(string)$select);
 			$htmlOptions['value']=$value;
 			$htmlOptions['id']=$baseID.'_'.$id++;
 			$option=self::radioButton($name,$checked,$htmlOptions);
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -1168,7 +1168,7 @@ class CHtml
 
 		foreach($data as $value=>$labelTitle)
 		{
-			$checked=!empty($select) && !is_array($select) && !strcmp($value,$select) || is_array($select) && in_array($value,$select);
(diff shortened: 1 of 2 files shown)


framework/web/helpers/CHtml.php
```diff
@@ -1168,7 +1168,7 @@ public static function checkBoxList($name,$select,$data,$htmlOptions=array())
 
 		foreach($data as $value=>$labelTitle)
 		{
-			$checked=!is_array($select) && !strcmp($value,$select) || is_array($select) && in_array($value,$select);
+			$checked=!empty($select) && !is_array($select) && !strcmp($value,$select) || is_array($select) && in_array($value,$select);
```
marcovtwout, line 1171: I'm not sure this is safe: `empty()` will also discard any falsy value like "0". Also I'm not entirely sure null is not allowed as a valid key for $data.

If you're specifically want to catch the error from strcmp() you could probably replace that comparison with one of these:
```
($select !== null && !strcmp($value,$select))
!strcmp($value,(string)$select) // this is how it is solved elsewhere in this file, see CHtml::listOptions()
```

And could you also check radioButtonList()? It has the same `!strcmp($value,$select);` and might me missing a fix also.

marcovtwout (changes requested):

framework/web/helpers/CHtml.php
rob006: I think that `strcmp($value,(string)$select)` is the way to go - it worked like that previously, but cast to string was implicit.

csears123 (commented):
I think the better solution would be to cast the argument in strcmp() as a string.  As well as the same solution for radioButtonList() method

framework/web/helpers/CHtml.php
```diff
@@ -1168,7 +1168,7 @@ public static function checkBoxList($name,$select,$data,$htmlOptions=array())
 
 		foreach($data as $value=>$labelTitle)
 		{
-			$checked=!is_array($select) && !strcmp($value,$select) || is_array($select) && in_array($value,$select);
+			$checked=!is_array($select) && !strcmp($value, (string)$select) || is_array($select) && in_array($value,$select);
```
marcovtwout, line 1171: Nitpick: yii coding style has no space between arguments.

marcovtwout (approved):
Change looks good now.

You may also add your name to this line of the changelog: https://github.com/yiisoft/yii/blob/master/CHANGELOG#L7

framework/web/helpers/CHtml.php
marcovtwout: @csears123 It doesn't make sense to use new coding standards only for bits of new code - common practise is to stick to the framework conventions unless you are doing very large changes.

Arhell (approved):

PR #4495, opened 2022-11-18
Optimize PNG files

dilyanpalauzov:
by calling `for i in $(find -name '*.png'); do optipng -o7 -zm9 $i; done`.


samdark (approved):

samdark:
Thanks.

marcovtwout:
@dilyanpalauzov Could you tell me why we would not use the officially distributed files and optimize these?

samdark:
@marcovtwout just these are a bit smaller :) I did that in the past but it was lost during one of the updates.

marcovtwout:
@samdark Thanks for clearning that up. One might otherwise wonder why vendor files are modified / will give a binary signature mismatch. To save just a few hundred bytes I would personally use the originals, but if this was common practice before in yii1 I guess it makes sense.

dilyanpalauzov:
I optimized the PNG files in Limesurvey.  Then I was told that once upstream synchronization happens, the changes will be reverted.  In turn I optimized the PNG files used upstream projects, like here or https://github.com/jquery/jquery-ui/pull/2132.

PR #4386, opened 2021-12-01
PHP 8.1 support

marcovtwout:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4385

List of issues:
- [x] Fix tests on Composer v2 and PHP 8.0+ through https://github.com/yiisoft/phpunit-patches/pull/8
- [X] Return type of (..) should either be compatible with (..) void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice (https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.core.type-compatibility-internal)
- [x] Various cases of https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal
- [x] Various cases of https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.implicit-float-conversion
- [x] Wait for HTMLPurifier PHP 8.1 support in next release 4.15.0 https://github.com/ezyang/htmlpurifier/issues/311 / https://github.com/ezyang/htmlpurifier/pull/310
- [X] PHP 8.1 Cast decimals to string on sqlite https://github.com/yiisoft/yii2/issues/19144

Please list additional issues found in https://github.com/yiisoft/yii/issues/4385

PR #4500, opened 2022-12-12
PHP 8.1 deprecation error, add return type for CMssql-classes

csears123:
The return type for lastInsertId() method needs to be compatible with PDO::lastInsertId().  Adding the return type for stiring|false resolves the compatibility issue encountered in PHP 8.1

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,7 @@
 Version 1.1.28 under development
 --------------------------------
 
-- No changes yet
+- Bug #4500: PHP 8.1 compatibility: Fix deprecation warnings in CMysql classes (csears123)
 
 Version 1.1.27 November 21, 2022
 --------------------------------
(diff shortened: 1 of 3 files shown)


marcovtwout:
@csears123 Please note that code needs to stay compatible with older PHP versions. For alternative ways to solve you could take a look at solutions applied in https://github.com/yiisoft/yii/pull/4386

marcovtwout:
@csears123 Similar fixes are required in `CMssqlPdoAdapter`, which is the only other class that extends `PDO`.

marcovtwout (approved):

marcovtwout:
Thanks! Merging.

Issue #4491, opened 2022-10-24
Limit and Offset not working correctly with MSSQL


Shnoulle:
### What steps will reproduce the problem?

Use an (MS) SQL server up to 11
get a data table of 150 lines in SQL server
Create a SQL request on a big table and add limit of 100 and offset of 100

```
$criteria = new CDBCriteria();
$criteria->order = 'id';
$criteria->limit = 100;
$criteria->offset = 100;

$Lines = MyModel::model()->findAll($criteria);
echo count($Lines);
```

Real code : 
https://github.com/LimeSurvey/LimeSurvey/blob/5f7030cc4cdc15bb2ec35768f933d50dbe048096/application/helpers/admin/exportresults_helper.php#L115
Can be seen too on Grid 
https://bugs.limesurvey.org/view.php?id=18420
### What is the expected result?

get 50 lines

### What do you get instead?

get 100 lines

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.26
| PHP version      | 7.4
| Operating system | Windows IIS
| SQL Server | SQLServerVersion 15.00.4249

- **Related YII2 fix** : https://github.com/yiisoft/yii2/blob/364e907875fd57ee218085cca796ac5d1c3c8d51/framework/db/mssql/QueryBuilder.php#L73
- **Possible Yii1 fix** : https://github.com/LimeSurvey/LimeSurvey/pull/2673

If OK : i create the pull request, i know it's not related to PHP version, but to SQL version.

marcovtwout:
Could you clarify: Is this a forward compatibility issue with newer MSSQL versions? Or a bug when using older versions of MSSQL? And which versions exactly are working/broken?

This seems to be the related issue for Yii 2: https://github.com/yiisoft/yii2/pull/4254, and the fix: https://github.com/yiisoft/yii2/commit/212c5ee3ef2eb244f7ca346ea1825427200b3394

Shnoulle:
> Could you clarify: Is this a forward compatibility issue with newer MSSQL versions? Or a bug when using older versions of MSSQL? And which versions exactly are working/broken?

It's forward compatibility issue with newer MSSQL. I'm sure it broke with SQL15, the related fix in Yii2 show SQL11

It's already fixed in Yii2, but not in Yii1 :)

Unsure MSSQL issue are OK to be merged.

marcovtwout:
[...]

PR #4501, opened 2022-12-13
Fix limit and offset not working correctly with MSSQL

wtommyw:
Generate the correct SQL for MSSQL 11 (2012) and newer, see #4491.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4491

### DIFF
--- a/framework/db/schema/mssql/CMssqlCommandBuilder.php
+++ b/framework/db/schema/mssql/CMssqlCommandBuilder.php
@@ -149,6 +149,32 @@ class CMssqlCommandBuilder extends CDbCommandBuilder
 	}
 
 	/**
+	 * Apply limit and offset to sql query
+	 * @param string $sql SQL query string.
+	 * @param integer $limit maximum number of rows, -1 to ignore limit.
+	 * @param integer $offset row offset, -1 to ignore offset.
+	 * @return string SQL with limit and offset.
+     * @see https://github.com/yiisoft/yii/issues/4491
+	 */
+	public function applyLimit($sql, $limit, $offset)
+	{
+		$limit = $limit!==null ? (int)$limit : -1;
+		$offset = $offset!==null ? (int)$offset : -1;
+
+		if($limit <= 0 && $offset <=0) // no limit, no offset
+			return $sql;
+		if($limit > 0 && $offset <= 0) // only limit
+			return preg_replace('/^([\s(])*SELECT( DISTINCT)?(?!\s*TOP\s*\()/i',"\\1SELECT\\2 TOP $limit", $sql);
+
+		if(version_compare($this->dbConnection->getServerVersion(), '11', '<'))
+			return $this->oldRewriteLimitOffsetSql($sql, $limit, $offset);
+		else
+			return $this->newRewriteLimitOffsetSql($sql, $limit, $offset);
+	}
+
+	/**
+	 * Rewrite sql to apply $limit and $offset for MSSQL database version 10 (2008) and lower.
+	 *
 	 * This is a port from Prado Framework.
 	 *
 	 * Overrides parent implementation. Alters the sql to apply $limit and $offset.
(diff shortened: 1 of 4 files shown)


tests/framework/db/schema/CMssqlCommandBuilderTest.php
```diff
@@ -0,0 +1,92 @@
...
+	/**
+	 * Mock a CDbConection, with a CMssqlSchema and CMssqlTableSchema
+	 */
+	public function setUp()
+	{
+		/*
+		 * Disable the constructor and mock `open` so that CDbConnection does not
+		 * try to make a connection
+		 */
+		$this->db = $this->getMockBuilder(CDbConnection::class)
```
marcovtwout, line 19: @wtommyw `::class` is not available in older PHP versions, see the failed CI test. Classname as string should work.

framework/db/schema/mssql/CMssqlCommandBuilder.php
```diff
@@ -204,7 +204,22 @@ public function applyLimit($sql, $limit, $offset)
...
 	/**
-	 * Rewrite sql to apply $limit > and $offset > 0 for MSSQL database.
+	 * @param string $sql sql query
+	 * @param integer $limit $limit > 0
+	 * @param integer $offset $offset > 0
+	 * @return string modified sql query applied with limit and offset.
+	 */
+	protected function rewriteLimitOffsetSql($sql, $limit, $offset)
+	{
+		if (version_compare($this->dbConnection->getServerVersion(), '11', '<')) {
```
marcovtwout, line 214: Yii 1 framework uses a bit older formatting standard, such as no space after if and no brackets for single line statements. Could you compare to other framework code and adjust accordingly?

framework/db/schema/mssql/CMssqlCommandBuilder.php
```diff
@@ -224,6 +239,27 @@ protected function rewriteLimitOffsetSql($sql, $limit, $offset)
...
+	 * @return string modified sql query applied w th limit and offset.
+	 */
+	protected function newRewriteLimitOffsetSql($sql, $limit, $offset)
+	{
+		$order = $this->findOrdering($sql);
+		if (empty($order)) {
+			$sql .= " ORDER BY (SELECT NULL)";
+		}
+
+		$sql .= sprintf(" OFFSET %d ROWS FETCH NEXT %d ROWS ONLY", $offset, $limit);
```
marcovtwout, line 258: This code is more compact and omits some conditions used in https://github.com/yiisoft/yii2/blob/364e907875fd57ee218085cca796ac5d1c3c8d51/framework/db/mssql/QueryBuilder.php#L88, is that on purpose?

tests/framework/db/schema/CMssqlCommandBuilderTest.php
```diff
@@ -0,0 +1,92 @@
...
+	{
+		$this->db->method('getServerVersion')->willReturn('10');
+
+		$command = $this->createFindCommand([
+			'select'=>'id, title',
+			'order'=>'title',
+			'limit'=>2,
+			'offset'=>3
+		]);
+
```
marcovtwout, line 49: Nitpick: in this file, you could omit the newlines between $command and $this->assert to make it clear those belong together.

marcovtwout (commented):
Thanks for your contribution! A few questions in my review.

Shnoulle:
SQLServerVersion 15.00.4261

- fix https://bugs.limesurvey.org/view.php?id=18547
- fix https://bugs.limesurvey.org/view.php?id=18420

framework/db/schema/mssql/CMssqlCommandBuilder.php
wtommyw: This function is being called from: https://github.com/yiisoft/yii/blob/8530c30959d5f68c0d1b0d611e3ce2bf32a8371a/framework/db/schema/mssql/CMssqlCommandBuilder.php#L195

If we look inside `applyLimit` we can see that `rewriteLimitOffsetSql` is only called when both `limit` and `offset` are set:
```php
if ($limit > 0 && $offset <= 0) //just limit
	$sql = preg_replace('/^([\s(])*SELECT( DISTINCT)?(?!\s*TOP\s*\()/i',"\\1SELECT\\2 TOP $limit", $sql);
elseif($limit > 0 && $offset > 0) // Limit and offset
	$sql = $this->rewriteLimitOffsetSql($sql, $limit,$offset);
```
As you can see the 'only offset' case is not handled, Causing `rewriteLimitOffsetSql` to only handle both limit and offset as opposed to https://github.com/yiisoft/yii2/blob/364e907875fd57ee218085cca796ac5d1c3c8d51/framework/db/mssql/QueryBuilder.php#L88, where it is possible for `limit` to be left out.

It turns out that querries only containing an `offset` were never generated for MSSQL. This could be fixed the same way as limit and offset have been fixed: by adding a `oldRewriteOffsetSql` and `newRewriteoffsetSql`, leaving `oldRewriteLimitOffsetSql` untouched.

 `oldRewriteOffsetSql($sql, $offset)` would generate the following SQL: 
```sql
SELECT *
FROM (
	SELECT *, 
	ROW_NUMBER() OVER (ORDER BY id) AS [__rowNumber__] -- Or order by (SELECT NULL) when no orderBy is given
	FROM [dbo].[item] [t]
) [t]
WHERE [__rowNumber__] > 15 -- Offset
```

And `newRewriteOffsetSql($sql, $offset)` would generate the following SQL
```sql
SELECT * FROM [dbo].[item]
ORDER BY id -- Or SELECT NULL) when no ordering is used
OFFSET 15 ROWS -- Offset
```

wtommyw:
@marcovtwout I've applied your feedback.

Offset wihtout limit has never worked, as described in [this comment](https://github.com/yiisoft/yii/pull/4501#discussion_r1048483419). I implemented this for MSSQL version 11 and newer.

marcovtwout (approved):

marcovtwout:
@Shnoulle Could you confirmed the updated version of the PR still works for your use case? After that I will merge it.

Shnoulle:
> @Shnoulle Could you confirmed the updated version of the PR still works for your use case? After that I will merge it.

I confirm : fix the 3 issue i found in LimeSurvey :3.X and 5.X version (removing temporary patch)

Shnoulle:
Great thank you to @wtommyw and @marcovtwout  :)

PR #4503, opened 2022-12-21
Fix PHP 8.1 deprecation errors

wtommyw:
Re-enabled deprecation errors in tests and fixed them.

Main fixes:
- CStarRating widget throwing a deprecation error if it is initalized without `value`;
- `strcmp`, `strtolower`, `mb_stringtolower`, `mb_strpos` and other built in functions being passed  `null` in tests;
- CFileValidatorTest testing for the wrong value, it returns 0 as the outcome is always cased to an int. The test never failed because `assertEquals` uses `==` and performs type coercion;
- CStringValidator treating `null` as an empty string, `strlen(null)` is deprecated so a cast is needed;
- ClientScript tests using a reference to an uninitialised variable, for example: https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/tests/framework/web/widgets/CAutoCompleteTest.php#L31 causing a deprecation error in: https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/framework/web/CClientScript.php#L414 as `$output` would be `null`.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4506

### DIFF
--- a/framework/i18n/CLocale.php
+++ b/framework/i18n/CLocale.php
@@ -113,7 +113,7 @@ class CLocale extends CComponent
 	 */
 	public static function getCanonicalID($id)
 	{
-		return strtolower(str_replace('-','_',(string)$id));
+		return strtolower(str_replace('-','_',$id));
 	}
 
 	/**
@@ -417,7 +417,7 @@ class CLocale extends CComponent
 	 */
 	public function getLocaleDisplayName($id=null, $category='languages')
 	{
-		$id = $this->getCanonicalID($id);
+		$id = $this->getCanonicalID((string)$id);
 		if (($category == 'languages') && (isset($this->_data[$category][$id])))
 		{
 			return $this->_data[$category][$id];
--- a/framework/i18n/CLocale.php
+++ b/framework/i18n/CLocale.php
@@ -113,7 +113,7 @@ class CLocale extends CComponent
 	 */
 	public static function getCanonicalID($id)
 	{
-		return strtolower(str_replace('-','_',$id));
+		return strtolower(str_replace('-','_',(string)$id));
 	}
 
 	/**
--- a/framework/utils/CDateTimeParser.php
+++ b/framework/utils/CDateTimeParser.php
@@ -83,9 +83,6 @@ class CDateTimeParser
 		if(self::$_mbstringAvailable===null)
 			self::$_mbstringAvailable=extension_loaded('mbstring');
 
-		if($value===null)
-			return false;
-
 		$tokens=self::tokenize($pattern);
 		$i=0;
 		$n=self::$_mbstringAvailable ? mb_strlen($value,Yii::app()->charset) : strlen($value);
--- a/framework/utils/CDateTimeParser.php
+++ b/framework/utils/CDateTimeParser.php
@@ -83,6 +83,9 @@ class CDateTimeParser
 		if(self::$_mbstringAvailable===null)
 			self::$_mbstringAvailable=extension_loaded('mbstring');
 
+		if($value===null)
+			return false;
+
 		$tokens=self::tokenize($pattern);
 		$i=0;
 		$n=self::$_mbstringAvailable ? mb_strlen($value,Yii::app()->charset) : strlen($value);
--- a/framework/utils/CFormatter.php
+++ b/framework/utils/CFormatter.php
@@ -73,7 +73,7 @@ class CFormatter extends CApplicationComponent
 	 * They correspond to the number of digits after the decimal point, the character displayed as the decimal point
 	 * and the thousands separator character.
 	 */
-	public $numberFormat=array('decimals'=>null, 'decimalSeparator'=>null, 'thousandSeparator'=>null);
+	public $numberFormat=array('decimals'=>0, 'decimalSeparator'=>null, 'thousandSeparator'=>null);
 	/**
 	 * @var array the text to be displayed when formatting a boolean value. The first element corresponds
 	 * to the text display for false, the second element for true. Defaults to <code>array('No', 'Yes')</code>.
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -2370,7 +2370,7 @@ EOD;
 	public static function value($model,$attribute,$defaultValue=null)
 	{
 		if(is_scalar($attribute) || $attribute===null)
-			foreach(explode('.',$attribute) as $name)
+			foreach(explode('.',(string)$attribute) as $name)
 			{
 				if(is_object($model))
 				{
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -2369,8 +2369,10 @@ EOD;
 	 */
 	public static function value($model,$attribute,$defaultValue=null)
 	{
-		if(is_scalar($attribute) || $attribute===null)
-			foreach(explode('.',(string)$attribute) as $name)
+		if($attribute===null)
+			return $defaultValue;
+		elseif(is_scalar($attribute))
+			foreach(explode('.',$attribute) as $name)
 			{
 				if(is_object($model))
 				{
--- a/framework/web/helpers/CHtml.php
+++ b/framework/web/helpers/CHtml.php
@@ -2369,10 +2369,8 @@ EOD;
 	 */
 	public static function value($model,$attribute,$defaultValue=null)
 	{
-		if($attribute===null)
-			return $defaultValue;
-		elseif(is_scalar($attribute))
-			foreach(explode('.',$attribute) as $name)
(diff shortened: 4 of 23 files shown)


framework/i18n/CLocale.php
```diff
@@ -113,7 +113,7 @@ protected function __construct($id)
 	 */
 	public static function getCanonicalID($id)
 	{
-		return strtolower(str_replace('-','_',$id));
+		return strtolower(str_replace('-','_',(string)$id));
```
rob006, line 116: You should not pass `null` to this method according to phpdoc. In what scenario you will get deprecation error here?

framework/utils/CDateTimeParser.php
```diff
@@ -83,6 +83,9 @@ public static function parse($value,$pattern='MM/dd/yyyy',$defaults=array())
 		if(self::$_mbstringAvailable===null)
 			self::$_mbstringAvailable=extension_loaded('mbstring');
 
+		if($value===null)
```
rob006, line 86: `$value` needs to be string according to phpdoc. I think it makes more sense to handle null/empty string directly in validators instead of here.

framework/web/helpers/CHtml.php
```diff
@@ -2369,7 +2369,9 @@ public static function listData($models,$valueField,$textField,$groupField='')
 	 */
 	public static function value($model,$attribute,$defaultValue=null)
 	{
-		if(is_scalar($attribute) || $attribute===null)
+		if($attribute===null)
```
rob006, line 2372: `null` could be a valid key for an array, returning `$defaultValue` blindly is incorrect.

tests/bootstrap.php
```diff
@@ -2,7 +2,7 @@
 
 if (PHP_MAJOR_VERSION>=7 && PHP_MINOR_VERSION>=1) {
 	// skip deprecation errors in PHP 7.1 and above
-	error_reporting(E_ALL & ~E_DEPRECATED);
+	error_reporting(E_ALL);
```
rob006, line 5: :+1:

rob006 (changes requested):

framework/i18n/CLocale.php
wtommyw: > You should not pass null to this method according to phpdoc

I agree, it would probably be better to change the test passing `null` to it in the first place. We could either:
**A:** Cast the provided data to a string, for example:
https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/tests/framework/i18n/CLocaleTest.php#L126
Would become:
```php
 $this->assertEquals(mb_strtolower((string)$assertion),mb_strtolower((string)$locale->getLocaleDisplayName((string)$methodLocale))); 
 ```
Note that the first two casts are required as it is possible for [`getLocaleDisplayName`](https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/framework/i18n/CLocale.php#L410-L418) to return `null` into `mb_strtolower`, which has been deprecated.

Or **B:** not pass `null` into `getLocaleDisplayName` at all, as it isn't allowed according to the phpdoc. This would mean that the dataProviders should be changed, for example:
https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/tests/framework/i18n/CLocaleTest.php#L107-L108
Could be changed to something like:
```php
 array('es_MX','',null),  // Empty string, essentially the same as (string)null
 array('es_ES','',null),  // Or any invalid ID causing getLocaleDisplayName to return null
```

@rob006 what are your thoughts on this? I think option **B** might be somewhat clearer for anyone reading the tests

rob006: I would remove these tests completely, they look invalid to me (they do not respect types from phpdoc). But maybe I'm missing something, so I would wait for second opinion on this.

framework/utils/CDateTimeParser.php
wtommyw: This is the same issue as https://github.com/yiisoft/yii/pull/4503#discussion_r1054552182

This is also a case where the tests are passing `null` and `false` into a method that only accepts `$value` as a string. https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/tests/framework/utils/CDateTimeParserTest.php#L196-L197

I looked for all the usages of `CDateTimeParser::parse` and it appers that `$value` is always checked for `null` before being passed to `CDateTimeParser::parse`: https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/framework/validators/CTypeValidator.php#L123-L128
and https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/framework/validators/CDateValidator.php#L52

These invalid assertions should be safe to remove but I'll wait for https://github.com/yiisoft/yii/pull/4503#discussion_r1054571980 in the event that I am missing something.

framework/i18n/CLocale.php
marcovtwout: If only tests code calls this function with `null` and never framework code, I would remove the tests. Otherwise, I would vote to keep current behavior and preventing possible regressions by casting to string.

wtommyw: The function defaults `$id` to null:
https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/framework/i18n/CLocale.php#L418

Which passes `$id` on to:
https://github.com/yiisoft/yii/blob/1c6f6fe221404de9afeb32059e6671676237482c/framework/i18n/CLocale.php#L108-L114
Which doesn't allow for `$id` to be null.

I think it might be a good idea to cast to string to prevent possible regressions, because `getLocaleDisplayName` is public and allows `$id` to be `null`. Bven though no framework functions call `getLocaleDisplayName` without parameters or with `null` . I'm not quite sure where the best place to cast would be:
1. In the test, although this would not cover `getLocaleDisplayName` without any parameters;
2. In `getLocaleDisplayName`, while passing it to `getCanonicalID`;
3. Or in `getCanonicalID` as it currently is in my fork.

marcovtwout: Since a public framework function allows $null and passes it on, I consider this a call from framework code. So I suggest to keep the current behavior by casting to string as you currently have it.

rob006: Accepting `null` in `getLocaleDisplayName()` looks nonsensical IMO - canonicalizing ID is useful only for fetching language name and it will always return `null` in that case (there is no name for "null-language"), so there is no point of calling `getCanonicalID()` at all. IMO we should not propagate such weirdness from `getLocaleDisplayName()` to `getCanonicalID()` and deal with `null` locally :stuck_out_tongue:.

wtommyw:
@marcovtwout @rob006 I tried to request a review on both of you, but it looks like Github does not allowe this. I'm unsure if anyon's got the ping now

rob006 (approved):
I still think that cast in `getCanonicalID` is unnecessary and `null` should be handled in `getLocaleDisplayName()`  https://github.com/yiisoft/yii/pull/4503/files#r1055466147

But the rest looks OK.

framework/i18n/CLocale.php
marcovtwout: @rob006 I agree it's not nice from a design point of view and an inconcistency, trivial as it may be. @wtommyw I'm fine with either solution and leave the final decision up to you. :)

marcovtwout:
@wtommyw After https://github.com/yiisoft/yii/pull/4503/files#r1055466147, feel free to add a changelog line with something like : "Enh 4503: Update tests/code for PHP 8.1 compatibility and stop suppressing deprecation warnings"

wtommyw:
> I still think that cast in `getCanonicalID` is unnecessary and `null` should be handled in `getLocaleDisplayName()` https://github.com/yiisoft/yii/pull/4503/files#r1055466147

I agree, I've made the change. This way we'll actually follow the documented behaviour (not passing null into `getCanonicalID`.

framework/utils/CFormatter.php
```diff
@@ -73,7 +73,7 @@ class CFormatter extends CApplicationComponent
 	 * They correspond to the number of digits after the decimal point, the character displayed as the decimal point
 	 * and the thousands separator character.
 	 */
-	public $numberFormat=array('decimals'=>null, 'decimalSeparator'=>null, 'thousandSeparator'=>null);
+	public $numberFormat=array('decimals'=>0, 'decimalSeparator'=>null, 'thousandSeparator'=>null);
```
marcovtwout, line 76: 👍 in line with PHP docs for number_format()

marcovtwout (approved):

PR #4515, opened 2023-02-16
PHP 8.1 compatibility: Fix deprecation warning in CHtml::value 

papppeter:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | Explode cannot accept null.


marcovtwout:
@papppeter You should be able to pass $attribute as `null` to CHtml::value, as it is cast to string in explode. Did you really test on the latest Yii version?

papppeter:
> @papppeter You should be able to pass $attribute as `null` to CHtml::value, as it is cast to string in explode. Did you really test on the latest Yii version?

thanks for geting back so quickly. No, i was just finishing the test. I just recognized that master has already a fix for this issue.

PR #4520, opened 2023-03-23
Update YiiBase.php to fix autoload to handle null includePaths

ganey:
Fix static autoload to handle case where _includePaths isn't set yet (so isn't an array)

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌

### DIFF
--- a/framework/YiiBase.php
+++ b/framework/YiiBase.php
@@ -444,7 +444,7 @@ class YiiBase
 			{
 				if(self::$enableIncludePath===false)
 				{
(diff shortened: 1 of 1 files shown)


marcovtwout:
@ganey Could you supply a simple use case where this happens? Doing foreach() over null should have thrown a PHP warning for a long time already.

marcovtwout:
Closing for inactivity.

PR #4524, opened 2023-06-26
CDbTransaction raises "commit" and "rollback" events.

humeniuc:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Problem: 
In a nested transaction some operations should be executed only after the transaction was successfully committed.  
There is no way to postpone the execution of some code until after the transaction is committed.

Solution: 
The transaction object triggers `onCommit`/`onRollback` events.  
One could register an listener to the transaction object which is executed only if the transaction is successfully committed.

### DIFF
--- a/framework/db/CDbTransaction.php
+++ b/framework/db/CDbTransaction.php
@@ -62,7 +62,8 @@ class CDbTransaction extends CComponent
 		{
 			Yii::trace('Committing transaction','system.db.CDbTransaction');
 			if($this->_connection->getPdoInstance()->inTransaction())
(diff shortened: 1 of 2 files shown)


marcovtwout:
While the addition looks simple enough, Yii is in maintainance mode and currently only accepting security and compatibility fixes. Also it differs a bit from the event handling that Yii2's database transaction already has, so could be confusing to add this in another way.

PR #4528, opened 2023-08-08
fix PHP 8.2 error "Passing null to parameter is deprecated"

lubosdz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Output buffer(s) may also be empty (containing NULL), which throws an error in PHP 8.2+.

### DIFF
--- a/framework/web/CClientScript.php
+++ b/framework/web/CClientScript.php
@@ -411,7 +411,7 @@ class CClientScript extends CApplicationComponent
 		if($html!=='')
 		{
 			$count=0;
-			$output=preg_replace('/(<title\b[^>]*>|<\\/head\s*>)/is','<###head###>$1',$output,1,$count);
+			$output=preg_replace('/(<title\b[^>]*>|<\\/head\s*>)/is','<###head###>$1',(string)$output,1,$count);
 			if($count)
 				$output=str_replace('<###head###>',$html,$output);
 			else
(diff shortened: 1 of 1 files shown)


rob006:
> Output buffer(s) may also be empty (containing NULL),

In what scenario it will be `null`?

lubosdz:
@rob006 I really don't know. I was testing older Yii.1 online app on PHP 8.2.3 to see how costy it would to to upgrade. Out of all traced issues there was the only one caused by framework (Yii 1.1.28) - on line [414](https://github.com/yiisoft/yii/blob/30899af7b84b050a47f44a26959c25f0e33c8526/framework/web/CClientScript.php#L414) (method `CClientScript::renderHead()`). Casting to `string` fixed the issue. Then I noted same `preg_match()` expressions in `renderBodyBegin()` and `renderBodyEnd()` and added to PR to potentially fix those too. I am not going to investigate details why was buffer `NULL` on line 414, no time for that. Feel free to close this PR if you think it's not justified or wait till someone confirms.

rob006:
These methods do not accept nulls according to phpdoc, so if we want to handle nulls here, we should update phpdoc too. But IMO it would be better to identify case when this is happening and ensure that output is string before we pass it to methods for output processing.

PR #4529, opened 2023-08-09
Show cause of the error when fixture can't not be loaded

eduardor2k:
When a fixture can't be loaded, you only get a generic error telling you which fixture can't be loaded, it would be more helpful if in that message we get the cause of the error

`1) Unit\RequestApiTest::testCreate
CException: Exception loading row req1 in fixture /var/www/html/test/yiitools/protected/tests/fixtures/requests.php` 

to

`1) Unit\RequestApiTest::testCreate
CException: Exception loading row req1 in fixture /var/www/html/test/yiitools/protected/tests/fixtures/requests.php Error: CDbCommand failed to execute the SQL statement: SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '1691576897' for column 'date_added' at row 1`

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 Version 1.1.29 under development
 --------------------------------
 
+- New: Improved error messaging in the database fixture manager file (eduardor2k)
 - Bug #4516: PHP 8 compatibility: Allow union types and intersection types in action declarations (wtommyw)
 - Bug #4523: Fixed translated in Greek class messages in framework requirements view, which they should not be translated (lourdas)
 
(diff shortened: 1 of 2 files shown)


marcovtwout (approved):
The original exception message is not lost (passed as 3rd argument to CException) but it might depend on the environment or error handler if you get to see the full stack trace containing that error.

The change looks helpful though and I see no BC issues. Could you add a line to the CHANGELOG as well?

marcovtwout:
Merged, thanks!

PR #4530, opened 2023-08-10
Update PHP supported version from 8.1 to 8.2

eduardor2k:
Update the docs to show that the current php version 8.2 is supported, showing 8.1 it might give the impression that yii does not support current versions of php.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️


marcovtwout (approved):
Good one!

PR #4531, opened 2023-08-29
#[\AllowDynamicProperties] in ActiveRecord Models

eduardor2k:
Set this class to Allow Dynamic Properties, this will allow all child classes to allow dynamic properties

https://www.php.net/manual/en/migration82.deprecated.php

This will prevent getting a warning trying to set a dynamic property in CActiveRecord objects

This can be fixed in userland code, but since this is the expected behaviour, I would expect not getting a warning when using dynamic properties.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/framework/db/ar/CActiveRecord.php
+++ b/framework/db/ar/CActiveRecord.php
@@ -37,6 +37,7 @@
  * @package system.db.ar
  * @since 1.0
  */
+#[\AllowDynamicProperties]
 abstract class CActiveRecord extends CModel
 {
 	const BELONGS_TO='CBelongsToRelation';


Arhell (commented):
@marcovtwout

rob006:
How do you get this warning? AR should already disallow dynamic properties with `__set()` implementation, and `AllowDynamicProperties` should not change anything.

rob006:
This is implemented by framework, so I doubt you will find any PHP docs regarding this:

https://github.com/yiisoft/yii/blob/d7845aa29dcc81846b5be35591781613a61244b5/framework/base/CComponent.php#L173-L174

PR #4532, opened 2023-08-29
Allow to instantiate the called class instead of the CActiveRecord class

eduardor2k:
The problem of this implementation is that this method needs to copied onto the children class to be instantiated correctly, since if the method is inherited _ _ CLASS _ _ will always point to CActiveRecord instead of the class calling this method, using get_called_class() fixes that, this function is present since php 5.3.0

Here's an example:

class A extends CActiveRecord {}

And we call A::model(), we will get an A instance instead of CActiveRecord

```php
<?php

class B
{
    public static function model($className = __CLASS__)
    {
        return new $className(null);
    }	
}

class A extends B
{
}

$b = new B();
$a = new A();

var_dump($b::model());
var_dump($a::model());
```

returns
```
object(B)#3 (0) { } 
object(B)#3 (0) { } 
```
but if we change that to get_called_class()

```php
<?php

class B
{
    public static function model($className = null)
    {
        if($className === null){
	     $className = get_called_class();
        }
        return new $className(null);
    }	
}

class A extends B
{
}

$b = new B();
$a = new A();

var_dump($b::model());
var_dump($a::model());
```
results in

```
object(B)#3 (0) { } 
object(A)#3 (0) { }
```

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

### DIFF
--- a/framework/db/ar/CActiveRecord.php
+++ b/framework/db/ar/CActiveRecord.php
@@ -388,8 +388,12 @@ abstract class CActiveRecord extends CModel
 	 * @param string $className active record class name.
 	 * @return static active record model instance.
 	 */
-	public static function model($className=__CLASS__)
+	public static function model($className=null)
 	{
+		if($className === null){
+			$className = get_called_class();
+		}
+		
 		if(isset(self::$_models[$className]))
 			return self::$_models[$className];
 		else


Arhell (commented):
@marcovtwout

twisted1919:
I don't think this should change, we've already accepted this in Yii1, we're all used to it, makes no sens to change now.

eduardor2k:
> I don't think this should change, we've already accepted this in Yii1, we're all used to it, makes no sens to change now.

you mean, having to clone the model method in every model, to get the correct instance? 

I'm sure you everybody has done this in all their yii1 projects, if true, this change, should be backwards compatible with those cases.

twisted1919:
Yeah, everybody had the copy of the method already because that's what Gii generates, that was my point, it is really a non-problem, all the models already have that when they are generated by Gii with the default template.   

What stops you to extend CActiveRecord, implement your change in that class, then extend your models from it?

eduardor2k:
@twisted1919 you're totally right, my reasoning is I don't see the point to keep it that way, having that code added to every model is pointless (literally), because the finality of the model() method is to get the instance of your model, not the CActiveRecord, it appears this comes from the days of yii 1.0, when get_called_class() did not exist

I would extend the scope of this, updating the model template and getting rid of the model method

twisted1919:
> I would extend the scope of this, updating the model template and getting rid of the model method  
plus updating the docs, I think the use case is heavy documented, I might be wrong though.  

I defer to @marcovtwout after all, whatever he thinks. It doesn't change things for me as long as there's no BC, if it makes your life easier, sure.

dimvic:
This is not a fully BC change.
If one has explicitly **not** included the method in their models to do something exotic, by exploiting the difference between `static` and `self` for example, this could be a breaking change.
Does not look like a real world scenario nor would that be good practice, but the case exists.

marcovtwout:
It's a nice idea, but for Yii 1 it wouldn't make sense anymore since the current implementation is well established and documented. Changing this now, even in a backward compatible manner, would only cause confusion.

PR #4544, opened 2024-01-29
fix/static method getCanonicalID should not be called dynamically

apphp:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌

### DIFF
--- a/framework/i18n/CLocale.php
+++ b/framework/i18n/CLocale.php
@@ -330,7 +330,7 @@ class CLocale extends CComponent
 	public function getLanguageID($id)
 	{
 		// normalize id
-		$id = $this->getCanonicalID($id);
+		$id = self::getCanonicalID($id);
 		// remove sub tags
 		if(($underscorePosition=strpos($id, '_'))!== false)
 		{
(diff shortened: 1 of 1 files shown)


marcovtwout (approved):

marcovtwout:
Thanks!

rob006:
This change breaks BC: previously if you extend this class and override `getCanonicalID()`, new implementation was used everywhere. After this change new implementation will be ignored and always `CLocale::getCanonicalID()` will be used.

marcovtwout:
@rob006 Thank you for checking in detail, but that is not exactly the case. The method itself was not updated and was already static, and there are more function calls within the framework that were already calling it correctly using `::getCanonicalID()`

rob006:
@marcovtwout 

Before change: https://3v4l.org/ZAIWn
After change: https://3v4l.org/bM5Ka

I'm not sure if overwriting `getCanonicalID()` is even supported, but `$this->` could be used here to emulate `static::` which was not available in older versions of PHP.

marcovtwout:
The old behavior (to my surprise) doesn't cause any PHP warning. Looking at the history, it doesn't look like $this was used on purpose, so I would say this change is fairly safe.

PR #4545, opened 2024-01-29
Method CDbCommandBuilder:: createMultipleInsertCommandWithIgnore added

apphp:
Added method CDbCommandBuilder:: createMultipleInsertCommandWithIgnore and implemented for MySQL and Postgres


marcovtwout:
Sorry, but Yii 1 Yii is in maintainace mode, only receiving security and compatibility fixes. This PR looks like optional functionality, which is not something we are taking into account right now.

apphp:
Get it, thanks!

PR #4546, opened 2024-01-30
FIX - Static method queryFormats should not be called dynamically

apphp:
Static method $imagick::queryFormats() should not be called dynamically

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,8 @@
 Version 1.1.30 under development
 --------------------------------
 
-- No changes yet
+- Bug #4541: PHP 8 compatibility: Fix deprecation in MarkdownParser (mdeweerd)
+- Bug #4544: PHP 8 compatibility: Fix deprecation in CLocale (apphp)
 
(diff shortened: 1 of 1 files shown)


apphp:
image

framework/web/widgets/captcha/CCaptcha.php
```diff
@@ -169,7 +169,7 @@ public static function checkRequirements($extension=null)
 		if(extension_loaded('imagick'))
 		{
 			$imagick=new Imagick();
-			$imagickFormats=$imagick->queryFormats('PNG');
+			$imagickFormats=$imagick::queryFormats('PNG');
```
marcovtwout, line 172: The `$imagick=new Imagick();` is no longer nessecary.

marcovtwout (changes requested):
Thanks for your contribution.

Your PR contains some irrelevant historical commits. Could you rebase and only supply the relevant commit, and also update CHANGELOG.md?

PR #4548, opened 2024-02-01
fix/static methods of HTMLPurifier should not be called dynamically

apphp:
Static methods of HTMLPurifier: escapeCommentedCDATA, escapeCDATA and removeIEConditional should not be called dynamically


marcovtwout:
HTMLPurifier is an external library, so this should be fixed upstream in https://github.com/ezyang/htmlpurifier
Also, since it's not technically a PHP warning/deprecation fixing it is optional.

PR #4550, opened 2024-03-02
Added the CDbCommand::throwExceptionIfSqlQueryTextExists() method

JustMisha:
This method is called by the methods that populate the $_query array to prevent hard to debug errors. 

If __construct() or setText sets $_text there is no point in using the methods like select(), from(), etc after that because the sql query text already exists and the buildQuery() method will never be called.

### DIFF
--- a/framework/db/CDbCommand.php
+++ b/framework/db/CDbCommand.php
@@ -151,9 +151,9 @@ class CDbCommand extends CComponent
 		return $this;
 	}
 
-	/**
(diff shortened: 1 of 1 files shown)


marcovtwout:
Thanks for your PR, but Yii 1 is only accepting compatibility and security fixes as it is in maintainance mode. Also I understand why you would want the framework to prevent this kind of misusage of the query builder, but this is not something we can change easily and keep backward compatible.

Issue #4552, opened 2024-04-10
PHP 8.3 Compatibility


sandippingle:
### What steps will reproduce the problem?

PHP Deprecated the following features/functions in their release of 8.3:

https://www.php.net/manual/en/migration83.deprecated.php

## Reflection
Calling ReflectionProperty::setValue() with only one parameter is deprecated. To set static properties, pass null as the first parameter.

Possibly impacted code: 

```
public function setLocaleDataPath($value)
	{
		$property=new ReflectionProperty($this->localeClass,'dataPath');
		$property->setValue($value);
	}
```

## SQLite3 
Using exceptions is now preferred, warnings will be removed in the future. Calling SQLite3::enableExceptions(false) will trigger a deprecation warning in this version.

How can we plan PHP 8.3 compatibility?

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 1.1.29
| PHP version      | 8.3
| Operating system | Linux

marcovtwout:
Regarding the reflection deprecation, that point seems valid. Those lines could probably be replaced with:
```
$class=new ReflectionClass($this->localeClass);
$class->setStaticPropertyValue('dataPath',$value);
```

Regarding SQLite3 enableExceptions(), I don't see that being used anywhere in framework code?

TheGithubDev:
PHP 8.3 introduces several deprecations that impact the current codebase. Specifically:
- Saner Increment/Decrement operators
- Changes in ReflectionProperty usage
- SQLite3 exception handling

#### Proposed Changes

1. **Saner Increment/Decrement Operators**
   - **Action**: Conduct a thorough review of the codebase to identify any usages of increment/decrement operators that rely on deprecated behaviors. Update the code to align with PHP 8.3's expectations.
   - **Impact**: Ensure no deprecated warnings are triggered by these operators.

2. **ReflectionProperty::setValue() Method**
   - **Action**: Modify the `setLocaleDataPath` method to comply with the new requirement of passing `null` as the first parameter when setting static properties.
   - **Implementation**:
     ```php
     public function setLocaleDataPath($value)
     {
         $property = new ReflectionProperty($this->localeClass, 'dataPath');
         if ($property->isStatic()) {
             $property->setValue(null, $value);
         } else {
             $property->setValue($this, $value);
         }
     }
     ```
   - **Impact**: Ensure the method works correctly with both static and non-static properties under PHP 8.3.

3. **SQLite3 Exception Handling**
[...]

PR #4553, opened 2024-06-04
PHP 8.3 compatibility

marcovtwout:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | https://github.com/yiisoft/yii/issues/4552


composer.json
```diff
@@ -108,7 +108,8 @@
 			"phpunit/phpunit": {
 				"Fix PHP 7 compatibility": "https://yiisoft.github.io/phpunit-patches/phpunit_php7.patch",
 				"Fix PHP 8 compatibility": "https://yiisoft.github.io/phpunit-patches/phpunit_php8.patch",
-				"Fix PHP 8.1 compatibility": "https://yiisoft.github.io/phpunit-patches/phpunit_php81.patch"
+				"Fix PHP 8.1 compatibility": "https://yiisoft.github.io/phpunit-patches/phpunit_php81.patch",
+				"Fix PHP 8.3 compatibility": "tests/phpunit_php83.patch"
```
samdark, line 112: Could be moved to `https://yiisoft.github.io/phpunit-patches/`.

marcovtwout: Ofcourse, but only if you think this is the right solution.

mlpassos:
Hey @marcovtwout ,

Thanks for your work on Yii and the PHP community and on this... do we have a date or estimate for releasing this 8.3 compatibility? We are about to test this.. we are already on 8.1 and migrating to 8.3 now.

o//

composer.json
marcovtwout: @samdark Could you comment on the above? I can then see how to finish and merge this PR.

samdark: Yes. I think it is worth moving there.

samdark:
@marcovtwout merged compatibility patch.

samdark (approved):

PR #4561, opened 2024-08-03
schemaCachingDuration is not working

albertborsos:
I realized the `CDbConnection.schemaCachingDuration` property has no effect.
We use this config only in staging and production, but after the deploy we get exceptions for missing attributes.

That is why I wrote this test to prove that `schemaCachingDuration` settings does not work.

You can run this test from the `tests` folder:

`php ../vendor/bin/phpunit framework/db/CDbConnectionTest.php --filter testSchemaCachingDuration`

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  |

### DIFF
--- a/tests/framework/db/CDbConnectionTest.php
+++ b/tests/framework/db/CDbConnectionTest.php
@@ -45,6 +45,34 @@ class CDbConnectionTest extends CTestCase
 		$this->assertTrue($db->isInitialized);
 	}
 
+	public function testSchemaCachingDuration($cachingDuration = 2)
+	{
+		$app = new TestApplication(array(
+			'id' => 'testApp',
+			'components' => array(
+				'db' => array(
+					'class' => 'CDbConnection',
+					'connectionString' => 'sqlite::memory:',
+					'schemaCachingDuration' => $cachingDuration,
+                ),
+				'cache' => array(
+					'class' => 'CFileCache',
+                ),
+			),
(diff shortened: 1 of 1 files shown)


Arhell (commented):
@marcovtwout

albertborsos:
I was not fully accurate. `schemaCachingDuration` has effect, because it caches the schema, but after the duration it does not invalidate the cached schema.

marcovtwout:
If I read the code correctly, the duration is used between requests. But within one request table data is always cached (in https://github.com/yiisoft/yii/blob/master/framework/db/schema/CDbSchema.php#L34 which is always checked first in https://github.com/yiisoft/yii/blob/master/framework/db/schema/CDbSchema.php#L74). Can you confirm this is the case for you?

If you need to force new schema data within one request, you can call schema->refresh() as you've probably already seen.

albertborsos:
thanks @marcovtwout , we have issues between requests. So you are right, my test cases are not correct. I will update it soon.

Since then I realized we use a custom cache which stores data in couchbase (correctly with expiration), but I have to investigate it more deeply.

For now, I close this ticket, since there is a lack of chance there is this bug in the framework for 10+ years.

Thanks.

PR #4566, opened 2024-09-24
PHP 8 compatibility: Fix passing null to string parameter parameter

apphp:
PHP8.1 compatibility fixed.
PHP Deprecated:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /opt/phishing/versions/5.3/web/protected/framework/console/CConsoleCommand.php on line 232

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  |

### DIFF
--- a/framework/console/CConsoleCommand.php
+++ b/framework/console/CConsoleCommand.php
@@ -229,8 +229,8 @@ abstract class CConsoleCommand extends CComponent
 		$params=array();	// unnamed parameters
 		foreach($args as $arg)
 		{
-            if (is_null($arg))
-                continue;
+			if (is_null($arg))
+				continue;
 
 			if(preg_match('/^--(\w+)(=(.*))?$/',$arg,$matches))  // an option
 			{
--- a/framework/console/CConsoleCommand.php
+++ b/framework/console/CConsoleCommand.php
@@ -229,6 +229,9 @@ abstract class CConsoleCommand extends CComponent
 		$params=array();	// unnamed parameters
 		foreach($args as $arg)
 		{
+            if (is_null($arg))
+                continue;
+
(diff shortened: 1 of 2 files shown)


framework/console/CConsoleCommand.php
```diff
@@ -229,6 +229,9 @@ protected function resolveRequest($args)
 		$params=array();	// unnamed parameters
 		foreach($args as $arg)
 		{
+            if (is_null($arg))
+                continue;
+
```
marcovtwout, line 234: Yii uses tabs instead of spaces.

marcovtwout (changes requested):
Could you add an example to reproduce the issue?

framework/console/CConsoleCommand.php
apphp: Fixed.

apphp: 2. To reproduce this issue, simply create your own command that inherits from ConsoleCommand and uses resolveRequest() for parsing arguments, like:
`[$action, $options, $args] = $this->resolveRequest($this->arguments);`
Then, when you write unit test for tis command, you may call it directly with wrong attributes (for negative testing), something like that:
```
$cmd = new OptimizeCommand(null, null);
$cmd->run(['css', '-s', null], false);
```
If one of arguments will be a null, you get an error: 
`PHP Deprecated: preg_match(): Passing null to parameter https://github.com/yiisoft/yii/issues/2 ($subject)`

marcovtwout: I don't think there normally is a reason you should be able to pass null to a CLI command argument, but the change is valid and keeps backward compatibility.

PR #4567, opened 2024-10-02
Provided a "username" property for Redis cache configurations

arfeen:
Added a "username" property for the Redis connection authentication. A username may be used other then "default".

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  |


Arhell (commented):
@marcovtwout Please take a look when you have time

rob006:
Are you sure this is compatible with older versions of Redis?

arfeen:
Before REDIS 6.0, there was not concept of username but a single password. I'm closing this request to update the code for older versions of REDIS.

PR #4586, opened 2025-03-27
CI tests using development ini

marcovtwout:
This makes it more clear what error reporting level is being used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,6 +110,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           extensions: ${{ env.extensions }}
+          ini-file: development
           ini-values: date.timezone='UTC'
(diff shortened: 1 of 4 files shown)


marcovtwout:
@albertborsos I fixed the deprecations in the test code, but starting from PHP 8.3, deprecations occur within PHPunit itself also. Perhaps they can be patched, but for now I suppress them entirely as was done before (https://github.com/yiisoft/yii/commit/4c1ef3973f13004f51b388cdb9591bec588c5e40)

albertborsos:
Thanks! That is fine, I also set this in my project. It was confusing for me, because the yii error handler said it is a "PHP error" so I tought it is deprecated, but it is just a warning. Thanks for your work!

samdark (approved):

Arhell (approved):

PR #4587, opened 2025-04-15
Add socket connection support to CRedisCache.php

mateusmetzker:
This PR updates the Redis connection configuration to use a Unix socket (e.g. /var/run/redis/redis.sock) instead of a hostname/port.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/framework/caching/CRedisCache.php
+++ b/framework/caching/CRedisCache.php
@@ -13,7 +13,7 @@
  *
  * CRedisCache needs to be configured with {@link hostname}, {@link port} and {@link database} of the server
  * to connect to. By default CRedisCache assumes there is a redis server running on localhost at
- * port 6379 and uses the database number 0.
+ * port 6379 and uses the database number 0. It also supports redis server using {@link unixSocket}.
  *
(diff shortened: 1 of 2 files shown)


framework/caching/CRedisCache.php
```diff
@@ -90,8 +108,10 @@ class CRedisCache extends CCache
 	 */
 	protected function connect()
 	{
+		$socket = $this->unixSocket ? 'unix://'.$this->unixSocket : $this->hostname.':'.$this->port;
```
marcovtwout, line 111: I think changing `$socket` to `$address` is more appropiate.

marcovtwout (changes requested):
This was attempted before with https://github.com/yiisoft/yii/pull/4469, but even though we are not officially accepting feature changes, your solution looks safe enough and more in line with Yii 2.

Could you also update the CHANGELOG?

Arhell (approved):

samdark (approved):

marcovtwout (approved):

PR #4589, opened 2025-04-17
refactor(security): add pure-PHP HMAC fallback

mdeweerd:
This fixes issues with deprecations in hash_hamac by adding a pure-PHP HMAC fallback implementation to handle cases where the native `hash_hmac` function for the algorithm is not available.

- Added `my_hash_hmac` function for pure-PHP HMAC calculation
- Added `hash_hmac_fallback` wrapper to use native `hash_hmac` if available, otherwise falls back to `my_hash_hmac`
- Updated `computeHMAC` to use the new fallback mechanism

After a PHP upgrade some feature relying on a specific hash_hmac algorithm did not work any more because it is removed from PHP.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 Version 1.1.32 under development
 --------------------------------
 
+- Bug #4590: CSecurityManager: add fallback for dropped hash_hmac algorithms (PHP7.2.0)
 - Enh #4587: Add socket connection support to `CRedisCache` (mateusmetzker)
 
 Version 1.1.31 April 10, 2025
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,7 @@
 Version 1.1.32 under development
 --------------------------------
 
-- Bug #4590: CSecurityManager: add fallback for dropped hash_hmac algorithms (PHP7.2.0)
+- Bug #4590: CSecurityManager: add fallback for dropped hash_hmac algorithms in PHP 7.2 (mdeweerd)
 - Enh #4587: Add socket connection support to `CRedisCache` (mateusmetzker)
 
 Version 1.1.31 April 10, 2025
--- a/framework/base/CSecurityManager.php
+++ b/framework/base/CSecurityManager.php
@@ -332,7 +332,7 @@ class CSecurityManager extends CApplicationComponent
 			$hashAlgorithm=$this->hashAlgorithm;
 
 		if(function_exists('hash_hmac'))
-			return hash_hmac($hashAlgorithm,$data,$key);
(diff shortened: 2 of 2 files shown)


CHANGELOG
```diff
@@ -4,6 +4,7 @@
 Version 1.1.32 under development
 --------------------------------
 
+- Bug #4590: CSecurityManager: add fallback for dropped hash_hmac algorithms (PHP7.2.0)
```
samdark, line 7:
```suggestion
- Bug #4590: CSecurityManager: add fallback for dropped hash_hmac algorithms in PHP 7.2 (mdeweerd)
```

framework/base/CSecurityManager.php
```diff
@@ -332,7 +332,7 @@ public function computeHMAC($data,$key=null,$hashAlgorithm=null)
 			$hashAlgorithm=$this->hashAlgorithm;
 
 		if(function_exists('hash_hmac'))
-			return hash_hmac($hashAlgorithm,$data,$key);
+			return self::hash_hmac_fallback($hashAlgorithm,$data,$key);
```
samdark, line 335: What are use-cases for using `computeHmac()` with sha1 or md5?

mdeweerd: Not sure what you mean by the question.  I encountered an issue when not using 'sha1' or 'md5' because hash_hmac stopped supporting it.

marcovtwout:
Can you explain why you want this feature added on the framework level? I don't see the framework itself using any of the algoritms that were disabled in PHP 7.2.0. (https://www.php.net/manual/en/function.hash-hmac.php#:~:text=7.2.0,joaat)%20was%20disabled.)

mdeweerd:
> Can you explain why you want this feature added on the framework level?

The framework is not setting other encodings internally, but the SecurityManager is used in a project with a different hashAlgorithm which is no longer supported natively.
Thing is that the result is out in the field on printed documents and the hash is used for a sanity check.

> My followup point would be that I don't think implementing our own fallback is a good idea from a security perspective, if you need such a thing perhaps a BC polyfill is available.

A security perspective depends on context.  SHA1 and MD5 are "broken" and still available.
BC - I could not find another solution, so I created this.

I keep a private fork of yii, so if this is not in the official branch that is not an issue.

marcovtwout:
@mdeweerd Your implementation is a fine solution for a private project, but from the framework perspective not required and more of an additional feature. I'll close the PR but it will be here if anyone runs into the same issue. Thanks for your contribution in any case :)

Issue #4578, opened 2025-01-23
PHP 8.4 Compatibility


marcovtwout:
PHP 8.4 was released last week: https://www.php.net/migration84
Please list any compatibility problems in this issue.

kevincobain2000:
Thanks @marcovtwout

marcovtwout:
Did anyone find more PHP 8.4 issues yet?

rob006:
> Here are a few deprecations we've found and fixed:

Are you sure about that? Things like `__construct(CDbConnection $connection,$query=null)` should work just fine.

jjdunn:
> > Here are a few deprecations we've found and fixed:
> 
> Are you sure about that? Things like `__construct(CDbConnection $connection,$query=null)` should work just fine.

good question.  maybe our syntax-linting is a bit too aggressive

see https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

re-reading this section, I think you're right:  if the type of the nullable parameter is not already declared, then it's probably ok.  So maybe these three fixes are unnecessary.  I'll check with the developer who is working on 8.4-compatibility in our application and see if he is actually seeing deprecation-warnings emitted in the logs...

samdark:
Seems all the tests we have pass well, so what's holding us from tagging another release?

samdark:
btw., adding Rector to pipeline might be a good idea. See:

1. https://github.com/yiisoft/db/blob/master/.github/workflows/rector-cs.yml
2. https://github.com/yiisoft/db/blob/master/rector.php

marcovtwout:
PRs merged. Please test on master!

PR #4602, opened 2025-12-03
Fix PHP 8.4 session_set_save_handler() deprecation

efhsg:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #4578

## Summary

Fix PHP 8.4 `session_set_save_handler()` deprecation warning when using custom session storage.

## Problem

PHP 8.4 deprecated the callback-style `session_set_save_handler()` with 6 individual callbacks. This will become an error in PHP 9.0.

## Solution

Added `CHttpSessionHandler` class implementing `SessionHandlerInterface` as an adapter that delegates to the existing `openSession`/`closeSession`/`readSession`/`writeSession`/`destroySession`/`gcSession` methods.

- PHP 8.0+: Uses object-style `session_set_save_handler()` with `CHttpSessionHandler`
- PHP < 8.0: Falls back to legacy callback-style (no deprecation warning exists there)

## Backwards Compatibility

Fully backwards compatible:
- Existing subclasses (`CDbHttpSession`, custom implementations) work without modification
- All overridden session handler methods continue to be called via delegation

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,7 @@ Version 1.1.32 under development
 --------------------------------
 
 - Enh #4587: Add socket connection support to `CRedisCache` (mateusmetzker)
+- Bug #4578: PHP 8.4 compatibility: Fix deprecation in session_set_save_handler() (efhsg)
 - Bug #4591: Fix deprecation in `CCaptchaAction` for PHP 8.1+ (rob006)
 
 Version 1.1.31 April 10, 2025


framework/web/CHttpSession.php
```diff
@@ -115,7 +115,28 @@ public function getUseCustomStorage()
 	public function open()
 	{
 		if($this->getUseCustomStorage())
-			@session_set_save_handler(array($this,'openSession'),array($this,'closeSession'),array($this,'readSession'),array($this,'writeSession'),array($this,'destroySession'),array($this,'gcSession'));
+		{
+			// PHP 8.4+ deprecates callback-style session_set_save_handler().
+			// Use object-style handler on PHP 7.0+ to avoid deprecation.
+			// CHttpSessionHandler is in a separate file to avoid parse errors on PHP 5.3
+			// where SessionHandlerInterface doesn't exist.
+			if(PHP_VERSION_ID >= 70000)
```
marcovtwout: Yii 1 uses this syntax by convention: `version_compare(PHP_VERSION,'7.0','>=')`

framework/web/CHttpSessionHandler.php
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	/**
+	 * @param int $max_lifetime
+	 * @return int|false
+	 */
+	#[ReturnTypeWillChange]
+	public function gc($max_lifetime)
+	{
+		return $this->_session->gcSession($max_lifetime) ? 0 : false;
```
marcovtwout: The [documentation on the interface](https://www.php.net/manual/en/sessionhandlerinterface.gc.php) says:

> Returns the number of deleted sessions on success, or [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) on failure. Note this value is returned internally to PHP for processing.

Since Yii doesn't count this number your implementation always returns int 0 on success. Do you expect any possible problems to arise from that? Maybe we could take a quick look in what way PHP wants to use this value for internal processing.

framework/web/CHttpSession.php
efhsg: We did try 8.0 and ran into failing tests, so I kept it at 7.0.
That feels like the best balance: the object-style handler is already stable on 7.x, and this keeps the behavior aligned across supported versions.

If we bump it to 8.0+, we’d intentionally keep PHP 7.x on the legacy callback path longer. That means two distinct runtime behaviors to maintain/test, and a bigger “version split” for apps still on 7.x.

PR #4608, opened 2026-01-05
fix: invalid path to /CHttpSessionHandler.php

tance77:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

This fixes version `1.1.32` if you are using Yii lite the path mapping was wrong.


Arhell (commented):
@marcovtwout

marcovtwout (changes requested):
Yiilite is generated so you cannot manually edit this file. The require_once is not even needed since core classes are found by the autoloader. I have another fix ready and will submit it soon.

PR #4614, opened 2026-02-24
Fix github builds

marcovtwout:
See commit message for details.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -116,6 +116,9 @@
 	"config": {
 		"allow-plugins": {
 			"cweagans/composer-patches": true
+		},
+		"audit": {
+			"block-insecure": false
 		}
 	}
 }


terabytesoftw:
I think it's better to add `--no-audit` to the composer line than to modify `composer.son`. 

https://github.com/yiisoft/yii/blob/a12f860b1b55ad5aea6349d851565835ebfde563/.github/workflows/build.yml#L136

marcovtwout:
@terabytesoftw We are limited in options since the older PHP version builds are using older composer versions (version 2.2.26 for PHP 5.3-7.1). Otherwise we could have used `--no-security-blocking` or `--no-audit`.

Edit: I just thought of the alternative env vars, that might work on all versions. I'll try this: https://getcomposer.org/doc/03-cli.md#composer-no-security-blocking

marcovtwout:
Yes, that works and composer.json can stay the same. Closing this PR, new PR: https://github.com/yiisoft/yii/pull/4615

PR #4615, opened 2026-02-24
Fix github builds

marcovtwout:
Replaces https://github.com/yiisoft/yii/pull/4614

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,7 +133,7 @@ jobs:
             php${{ matrix.php }}-composer-
 
       - name: Install dependencies with composer
-        run: composer update --no-interaction --no-progress --optimize-autoloader --ansi
+        run: COMPOSER_NO_SECURITY_BLOCKING=1 composer update --no-interaction --no-progress --optimize-autoloader --ansi
 
       - name: Run tests with phpunit
         run: |


.github/workflows/build.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             php${{ matrix.php }}-composer-
 
       - name: Install dependencies with composer
-        run: composer update --no-interaction --no-progress --optimize-autoloader --ansi
+        run: COMPOSER_NO_SECURITY_BLOCKING=1 composer update --no-interaction --no-progress --optimize-autoloader --ansi
```
samdark, line 136: I think it's better to be done in `composer.json` by adding `{"config": {"audit": {"block-insecure": false}}}`

marcovtwout: That was my previous solution: https://github.com/yiisoft/yii/pull/4614

But I prefer this since solution since we are limiting the security override to our github builds only. Anyone else building Yii in another context will still be informed of any security issues present and then make their own choice.

Or do you have another reasoning for putting it in composer.json?

terabytesoftw: If I think the developer should be informed about what they are installing, that will allow them to take precautions.

samdark: Nope. It's fine either way.

PR #4616, opened 2026-02-28
PDO_OCI returns NULL on INSERT...RETURNING operations in PHP 8.2+

andreaepifani-afk:
Forced $returnID to integer by initializing it to 0. This prevents a bug where PDO_OCI returns NULL on INSERT...RETURNING operations in PHP 8.2+. Tested with php 8.2.28

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️


marcovtwout:
This PR has 0 files changed? Could you also provide a reproducable example of the bug you encountered so we can verify this is a framework issue?

PR #4625, opened 2026-08-26
Fix PHP 8.6 deprecations

ac-dc:
## Summary
- replace deprecated `is_integer()` calls with `is_int()`
- add `create_sid()` and `validateId()` to the custom session handler
- add coverage for session ID generation and validation

## Tests
- targeted session tests: 3 tests, 89 assertions
- full suite has no new failures compared with baseline

### DIFF
--- a/framework/web/CHttpSessionHandler.php
+++ b/framework/web/CHttpSessionHandler.php
@@ -99,4 +99,28 @@ class CHttpSessionHandler implements SessionHandlerInterface
 	{
 		return $this->_session->gcSession($max_lifetime) ? 0 : false;
 	}
+
+	/**
+	 * Creates a new session ID.
+	 * @return string the new session ID
+	 */
+	public function create_sid()
+	{
+		if(function_exists('session_create_id'))
+			return session_create_id();
+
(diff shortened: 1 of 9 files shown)


framework/web/CHttpSessionHandler.php
```diff
@@ -99,4 +99,35 @@ public function gc($max_lifetime)
...
+	 * Returning true preserves the behavior of the legacy callback-based handler,
+	 * which accepted externally supplied session IDs.
+	 * @param string $id the session ID
+	 * @return bool whether the session ID is valid
+	 */
+	public function validateId($id)
+	{
+		return true;
+	}
 }
```
marcovtwout, line 126: What is the point behind this change? Please provide references to what changed in PHP.

And why is it mixed with a commit that changes is_integer() with is_int()?

ac-dc: My bad, closing this, and going to open separate PR's

ac-dc:
Thanks for the feedback. I split the unrelated PHP 8.6 deprecations into two focused PRs: session handler methods in #4626 and the is_integer() replacement in #4627.

